PR #5, opened 2015-10-28
Updated CSS scroll to auto

terra4383:
Attempt to fix weird scrolling issue in some browsers. Small change to css overflow.


SteveFeldman:
Looks good...

PR #8, opened 2015-10-29
Sidenav2

terra4383:
Fixed overflow/double scrollbar bug with side navigation in Windows


SteveFeldman:
Tested without issue...

PR #16, opened 2015-11-10
Colors

terra4383:
Slightly more muted colors to match universal styling choices as discussed by the UX team.


SteveFeldman:
Looks good

PR #17, opened 2015-11-12
Update manual contrast.lic install steps

robertzas:
Customer backed up his own contrast.lic to contrast.lic.bak with root permissions. TeamServer failed to start when the user created file could not be replaced (permissions) with TeamServer's contrast.lic.bak. 

Updated doc:
1. we do a license backup, the user doesn't explicitly need to backup the file themselves
2. what that file is called, so this particular issue shouldn't happen again


SteveFeldman:
Looks good...

PR #35, opened 2016-01-06
Update InstallingTeamServer.md

davewichers:
Various editorial enhancements, plus two fixes:
  1) Clarify that there is now only 1 URL for the login page for both user's and admins.
  2) Added a warning note for Linux users to make sure they've installed MySQL before proceeding with the install.


terra4383:
Thank you, we will get to this ASAP.

SteveFeldman:
Dave,

Thanks for the edits. Looks good...

Steve

PR #53, opened 2016-01-26
Admin teamserver updates

dhafley:
HOWTO Bring your own Tomcat and MySQL

### DIFF
--- a/content/admin/teamserver/config/DistributedConfiguration.md
+++ b/content/admin/teamserver/config/DistributedConfiguration.md
@@ -1,34 +1,50 @@
-# About Horizontal Scalability
-At this time, opting to scale your TeamServer horizontally should be considered only after you have worked with the Contrast Support Team to identify yourself as a candidate.  This HOWTO will guide you in the setup and configuration of additional software but please be aware you will be responsible for the monitoring and durability of additional software.  That being said, if you are familiar with installing and administering Tomcat and MySQL the process is straightforward to setup and maintain.
+# About Distributed Configuration
+As an enterprise-on-premise administrator, the typical way to install or update Contrast TeamServer is to log into [Hub](https://hub.contrastsecurity.com) and download the latest version for either your Linux or Windows operating system.  This artifact is an executable and includes all software to install, upgrade, and run Contrast TeamServer.  All software and storage is setup to run on one host.
 
-One of the first changes you will notice about this program is that the artifact delivered is a Java WAR file that you drop into your own Tomcat instance.  You cannot download this artifact from hub.contrastseurity.com currently but we are planning enhancements to Hub product for hosting this type of artifact.
+Distributing the components of TeamServer means that you are opting not to use some of the software bundled with Contrast TeamServer and will provide this software yourself.  
 
-We are interested in your progress!  If you are planning to explore scaling TeamServer horizontally, please contrast Contrast Support.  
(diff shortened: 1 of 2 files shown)


SteveFeldman:
@terra4383 let me be the primary reviewer on this one. Before we make this available, I have a number of other items we need covered in our documentation and with our Hub.

terra4383:
Ok will do

SteveFeldman:
David,

I have a few comments. It may make sense to discuss them before taking the next draft. I will put them in chronological order first ;)

1. The article should really be about "Distributed Configuration" versus horizontal scaling. Phase 1 of the documentation should be focused separating the container from the database. Rather, it reads as though we are immediately jumping to load-balanced configurations. I might not even discuss "load-balanced configurations" until a second or third version. Maybe you want a cool image, care of @terra4383 that shows the distributed server configuration?

Option 1: Looks like this is install EOP as is, then hook it up to an external DB.

Option 2: Looks like this involves installing own Tomcat 7, Java 7 and MySQL. Tomcat/Java could be on one server and MySQL on another?

2. We should try to refrain from specifically identifying Ubuntu. We should focus on Linux, referencing Ubuntu as our preferred OS platform. We support RHL, Centos, Ubuntu and Debian.

3. Let's move away from the phase "Bringing your own" and replace with something in the lines of "independently installed and managed" or something of the sort.

4. TeamServer presently runs MySQL 5.6.28

5. Do we have a sample my.cnf to recommend using?

6. I like the ansible setup you are offering. Would it make sense to simply link to our Github project for this?

7. So one thing the article assumes is that you have "already" installed TeamServer. That wasn't too clear to me. It may make sense to link to the EOP installation documentation very early in the document with a section about assumptions. 

Speaking of assumptions, we should probably add other items to this section about assumptions:

* Installation, configuration and administration of MySQL

* Installation, configuration and administration of Tomcat

* Installation, configuration and administration of Java

8. Do you want to provide any recommendations or commentary on sizing the database?

9. You may want to link up these two articles:
* https://docs.contrastsecurity.com/admin_tsconfig.html#backup
* https://docs.contrastsecurity.com/admin_tsconfig.html#encrypt

Comments below here are for Option 2:

10. This section assumes that a prior installation has been performed using the EOP installer?

11. I'm curious why you wouldn't need everything under $CONTRAST_HOME/data?  Maybe we recommend deleting the activemq directory, but I believe everything else is needed, unless the server can regenerate those directories after startup?

12. Why is .contrast and .initialized needed?

13. Would it make sense to have a section about storing logs on another drive for storage optimization?

14. Which version of Java are your recommending be installed?

15. The war they can grab from EOP should be called Contrast.war. Any reason they can't use that?

dhafley:
* Addresses vocabulary to refer to project as Contrast TeamServer Distributed Configuration (ctdc).  I've also pushed all code to [repo ctdc](https://github.com/Contrast-Security-OSS/ctdc).
* Address assumptions about current EOP installation in place, software versions, etc early in document.
* Use "Phase" instead of "Options", clarifying that this is a building process, not a decision tree.  
* Generalized operating system references to Linux and Windows.  All examples are clarified as examples for Ubuntu 14.04 for point of illustration.
* my.cnf and database size: I prefer not to go down this road as so much depends on host size and configuration.  I can work with customer to develop a my.cnf and sizing per deployment, if an internal dba is not available.
* The $CONTRAST_HOME directory includes the database, so I chose to single out what is important.  Everything else (rulepack, for example) is generated on TeamServer startup.
* .contrast and .initialized must be included to prevent license and esapi key rotation.  
* Clarified that Contrast.war can be brought from current EOP installation and that Contrast is exploring options to deliver this separately based on customer demand.

PR #59, opened 2016-02-04
CONTRAST-7244 updated with more recent properties

ubermitchy:
Added lots of properties.


SteveFeldman:
@ubermitchy you might also want to put a small note with code sample for generating the startup options. you had mentioned java -jar contrast.jar will show all of the options.

PR #66, opened 2016-02-16
Netdocs reorganization

terra4383:


content/user/agent/net/faq/DoesNotRecognizeThirdParty.md
```diff
@@ -0,0 +1,24 @@
+<!--
```
ldemidov: This should probably be removed.  We do identify third party .NET libraries now.

ldemidov: Nevermind, I updated the text to indicate that only Nuget packages are being identified now.

PR #76, opened 2016-03-03
Twostep

terra4383:
First draft...


SteveFeldman:
@terra4383 this is a fantastic read with great images. A couple comments/questions below:

1. Don't forget to swap out the image of the backup codes once we release a download backup code option.

2. We should hold off on "Forget device". Let's pull that out for now and update this doc down the road.

3. I made some edits directly...

PR #78, opened 2016-03-07
added tagging to the entire branch

SteveFeldman:
@jday3 and @terra4383 first pass at tags have been added...


jday3:
@SteveFeldman - Looks great - format is correct. Go ahead and merge

PR #115, opened 2016-06-20
Deleted code coverage page for Ticket #8826

nataliewiggins:


terra4383:
@nataliewiggins We don't want to delete articles altogether - we should archive them.  For now we'll leave this alone - since we are deleting the link to this article, it won't show up on the site.

PR #129, opened 2016-07-07
Quick Reference document around common options seen / used when confi…

dhafley:
Reference document to be used when assisting with SSL configurations.


SteveFeldman:
@dhafley I'm so sorry for the delay on this. It looks good from my perspective. You can merge.

dhafley:
Any feedback @meg23 ?

terra4383:
Once we hear from @meg23 I can create the new link for this and merge.

PR #132, opened 2016-07-19
Added a generateHAR.md and the gif that demonstrates the process.

donniepropst:


terra4383:
@donniepropst Steps 5 and 6 of your article (checking box next to "Preserve log" and clicking the "Clear" button) are not shown in the GIF. Can we update the image or was that intentional?

donniepropst:
I'll rerecord it and include those steps, and I was having an issue recording the developer tools window of Chrome with just a browser screen recorder but when I rerecord I'll see if I can find a way to just record Chrome.

PR #157, opened 2016-10-28
%20 apparently double encoded as %2520

palmergs:
In multiple places a URL encoded whitespace character (e.g. %20) looks like it was double encoded where the % character is converted after the space is converted (space => %20 then %20 => %2520).


terra4383:
Closing, as Galen has resubmitted to an internal branch within the repo.

PR #189, opened 2017-02-15
Contrast 12446 cf docs

donniepropst:
Added documentation for how to use the buildpack with a user provided service, how to setup a service broker to create service instances, and how to use the pivotal tile with the buildpack.


terra4383:
@SteveFeldman I will let you approve this doc but I would like to walk through it with @kaitlynmortimer tomorrow. We will also create the new link.

SteveFeldman:
Thanks @terra4383 for commenting. We definitely need the link(s).

PR #190, opened 2017-02-17
Contrast 12237

kaitlynmortimer:
New exclusions article


terra4383:
@SteveFeldman FYI this includes a corresponding branch with the same name on Bitbucket. Please just review the content if you don't mind, and I will push it after a final formatting review. Thanks!

terra4383:
@SteveFeldman The "URL" exclusion type is described here as "Applied URLs", which exists for the "Input" type as well. When you choose the "URL" type you only get the option for “These URLs”, not “All URLs” below. Do we need to rework the doc to reflect that a little better?

SteveFeldman:
@terra4383 yes...I think we should rework the content for accuracy.

PR #202, opened 2017-03-22
Contrast 12955 ldap certificates

kaitlynmortimer:
Copy edited existing doc and added certificates section.


kaitlynmortimer:
Note: I changed "bind" to "BIND" when used as a noun but kept it in lowercase when used as a verb.

PR #209, opened 2017-03-24
The "Where to find and Install" faq section is redundant

gilday:
The Eclipse section already has a page for installation which covers this topic


terra4383:
@gilday Thanks for this. I am closing this without merging because I went ahead and did it in a separate branch.

PR #220, opened 2017-04-13
Contrast 13399 clr2 references

kaitlynmortimer:
Copy edited and removed CLR2 references from three articles.


terra4383:
@kaitlynmortimer Looks fine to me.

PR #221, opened 2017-04-14
CONTRAST-13530-javascript-sdk-doc

donniepropst:
This is very similar to the Python SDK doc from a few weeks ago.  It will also be in a new folder under the SDK section.


kaitlynmortimer:
@donniepropst Can you add a brief intro - a few sentences is fine - to explain the What and Why?

PR #227, opened 2017-05-04
Od3 move images

miketmoore:
Removes non-content/design images.


terra4383:
Looks good you can merge!

PR #230, opened 2017-05-17
Contrast 14007 org stats revisions

kaitlynmortimer:
Updated text and screenshot in Apps section.


terra4383:
@kaitlynmortimer The image filename does not match to what's referred to in the article, therefore doesn't display. It's a capitalization issue - let's capitalize all image filenames please. Also, can you please take a new screenshot from the product rather than using the mockup? I personally like when the Technology chart is showing on the left for this screenshot, and we don't need tooltips showing. Thanks!

PR #231, opened 2017-05-19
Contrast 14083 node bluemix doc

kaitlynmortimer:
Based on draft in https://contrast.atlassian.net/browse/CONTRAST-14083


terra4383:
Looks good @kaitlynmortimer

PR #245, opened 2017-05-30
VSTS - More specific scope requirement

jday3:


kaitlynmortimer:
Do we want to add a note about potential requirements beyond the minimum (token)? Or would that info be overwhelming?

jday3:
@kaitlynmortimer I don't think that we need an additional comment about potential requirements. If we end up changing the requirements, there will be additional information in TeamServer that will mention that and we can update the docs at that time.

PR #248, opened 2017-06-01
updated screenshot and table

kaitlynmortimer:


content/admin/org_settings/integrations/JIRA.md
```diff
@@ -48,9 +48,14 @@ Continue to the fields below to set a vulnerability status based on each JIRA ti
```
ksrin: As an example, here are a few mappings that can be set for Jira status / resolution combinations.

PR #252, opened 2017-06-06
CONTRAST-14762. Fix versioning in docs for bluemix

cixel:
In general, I think that hardcoded version numbers should be avoid in general docs like this, because versions are subject to change so frequently.


kaitlynmortimer:
@cixel Based on your edits, should`0.9.0.tgz` be removed from the block in the Setup section?

cixel:
@kaitlynmortimer fixed

PR #279, opened 2017-08-01
Contrast 12932 quick start

palmergs:
1. Initial installation and setup information patterned after the structure of the Node agent.
2. Initial troubleshooting information focused on communication links between agent and service and between service and TeamServer


kaitlynmortimer:
@palmergs Editing is 99% finished. Please review my changes and make sure that I applied formatting correctly, particularly in the Installation article. (We use "`" for code and "*" for paths, file names, etc.)

PR #291, opened 2017-08-25
Remove documentation specific to the removed .NET profiler

DanFiedler:
Remove documentation specific to the profiler instrumentation mode that was removed in 3.4.3. Remove troubleshooting issues specific to profiler. Remove mention of different modes as there is only one mode now.

Note I deleted several files - not sure how to update the left hand menu or if I need to.


kaitlynmortimer:
@DanFiedler The articles that you'd like to remove completely should be put in an Archive folder. Do you still have copies of the files?

DanFiedler:
I don't but we should be able to pull them from git history.

PR #292, opened 2017-08-25
Level two rules update

ldemidov:
Updated the policy/level two documentation for latest changes on the agent.


kaitlynmortimer:
@ldemidov I'm 90% done copy editing these articles; but, I'm still a little confused about when it's accurate to use quotation marks versus backticks. Please correct my edits wherever necessary.

PR #310, opened 2017-09-22
Contrast 17305 library export update

kaitlynmortimer:


kaitlynmortimer:
@gilday - Please review the Track Libraries doc, and confirm that my explanation and example of using the API to get data are correct.

kaitlynmortimer:
@jday3 - Please review the About API doc, and confirm that my explanation of launching and using API documentation through the Contrast UI is accurate.

content/tools/api/AboutRestAPI.md
```diff
@@ -9,22 +9,29 @@ Contrast has a few API versions:
 * Application APIs (v2) 
 * HATEOS-compliant REST API (v3) **Recommended**
 
->**Note:** We haven't officially deprecated v1 or v2, but strongly recommend using v3.
+>**Note:** Contrast hasen't officially deprecated v1 or v2, but strongly recommend using v3.
```
terra4383: @kaitlynmortimer "hasn't"

content/tools/api/AboutRestAPI.md
```diff
@@ -9,22 +9,29 @@ Contrast has a few API versions:
...
+
+### Use the Documentation
+
+Go to the **API** and **Objects** sections to see a list of endpoints and related objects used by the endpoints. The **Flow** section provides a list of sample use cases and information on their sequence of endpoints that must be called to accomplish them. The documentation also includes a playground to test your endpoint interactions with v1, v2 and v3 API endpoints. This is a handy tool to quickly learn how to use the endpoints through experience. 
+
+To access the API documentation in the Contrast interface, authenticate with an administrator account in Contrast, go to the ** User menu > Organization Settings > API tab**, and click the button to **Launch Documentation**. Unlike the static Contrast API page, this version allows you to use actual data from your Contrast application to exercise the documentation.
```
terra4383: @jday3 Can you verify wording here?

content/user/apps/TrackLibraries.md
```diff
@@ -32,22 +32,36 @@ Export details on findings from the Libraries page, a library's overview page, o
 
 <a href="assets/images/Library-export.png" rel="lightbox" title="Export library details"><img class="thumbnail" src="assets/images/Library-export.png"/></a>
 
-In the Libraries grids, select individual libraries with the check boxes or use your filters to focus on specific data sets. When you apply a selection, you can choose to export all or selected libraries. If you don't select any checkboxes or filters, you can export data for all libraries. 
+In the Libraries grids, select individual libraries or use the filters to focus on specific data sets. If you apply a selection, you can choose to export all or selected libraries; if not, you can export data for all libraries.
```
terra4383: @kaitlynmortimer "grid" should be singular

content/user/apps/TrackLibraries.md
```diff
@@ -32,22 +32,36 @@ Export details on findings from the Libraries page, a library's overview page, o
...
+* Latest Version
+* Grade
+* SHA1
+* CVE Count
+* Application Count
+* Server Count
+* Number of Classes
+* Number of Used Classes
+
+You can access more details on your libraries by going to the [Contrast API documentation](https://api.contrastsecurity.com/#) **> Library > Libraries section**, or using the command line.
```
terra4383: The hope is that they are getting everything they need from our export, and I don't know how many additional details are actually available in all cases . So maybe we word this a little more subtly?  "Additional library details may be explored using the Contrast RESTful API documentation, or another manual method of your choosing: (then continue with the example below)

content/user/apps/TrackLibraries.md
```diff
@@ -32,22 +32,36 @@ Export details on findings from the Libraries page, a library's overview page, o
...
+* SHA1
+* CVE Count
+* Application Count
+* Server Count
+* Number of Classes
+* Number of Used Classes
+
+You can access more details on your libraries by going to the [Contrast API documentation](https://api.contrastsecurity.com/#) **> Library > Libraries section**, or using the command line. 
+
+> **Example:** To find related application names and export the data to a CSV, use a cURL request to the API:
```
terra4383: @gilday

content/tools/api/AboutRestAPI.md
```diff
@@ -9,22 +9,29 @@ Contrast has a few API versions:
...
 * HATEOS-compliant REST API (v3) **Recommended**
 
->**Note:** We haven't officially deprecated v1 or v2, but strongly recommend using v3.
+>**Note:** Contrast hasen't officially deprecated v1 or v2, but strongly recommend using v3.
 
 ## Overview
-The Contrast plugin allows you to collect nearly all the same information that is accessible through the Contrast UI. You can get vulnerability data, application, server and library information, download agents and much more.
 
-The API is available to both **SaaS** and **Enterprise On-Premises** users. The [endpoint for SaaS users](https://app.contrastsecurity.com/Contrast/api) is the same for everyone, while for On-Premise users it will depend on where the TeamServer application is installed on the client's network.
+The Contrast plugin allows you to collect nearly all the same information that is accessible through the Contrast interface - dowload agents, get vulnerability, application, server and library information, and much more. Contrast's goal is to make sure Contrast offers data to allow complete integration into your existing Continuous Integration (CI), SIEM software, or other business intelligence dashboards.
```
terra4383: @kaitlynmortimer typo: "download" - Also "Contrast's goal is to make sure Contrast..." sounds bad to me - can one of these switch? "Our goal is to make sure Contrast..." or "Contrast's goal is to make sure we..."

content/user/apps/TrackLibraries.md
terra4383: @kaitlynmortimer This second sentence just makes things sound confusing and complicated to me. How about "You may choose to either export all libraries or just those you've selected." I think the difference between what they can do from that icon with or without selections is easily discoverable in the UI and doesn't need to be so explicitly outlined.

content/tools/api/AboutRestAPI.md
gilday: "hasen't"

content/user/apps/TrackLibraries.md
gilday: It's unfortunate that our API Docs do not allow us to generate permalinks to the exact section. We should consider adding this

content/user/apps/TrackLibraries.md
gilday: > You can access more details on your libraries by going to the Contrast API documentation

This language sounds like you can access this data from the documentation, which might technically be true since our API Documentation is interactive and allows users to play with the API in their browser, but I would make the distinction that the data is available from the API, and users can learn about the API from the API Documentation

gilday: > or using the command line.

I would omit this because it sounds too prescriptive. We can offer an example which happens to use command line tools, but as written it sounds like there is a dedicated Contrast command line tool for access more details about libraries

gilday: > access more details

Should we specifically call out library associations to Applications and Servers? I believe this is the information that users are looking for a way to export because they want to create reports instead of viewing this data in Contrast

gilday: I will offer some language for what it's worth:

For users looking to craft custom software composition analysis reports about their applications, the library export feature may not provide sufficient information, but Contrast offers a rich Libraries API for accessing Contrast library data. Reference the [Contrast API documentation](https://api.contrastsecurity.com/#) **> Library > Libraries section** for instructions on using the Libraries API.

content/user/apps/TrackLibraries.md
```diff
@@ -32,22 +32,36 @@ Export details on findings from the Libraries page, a library's overview page, o
...
+* Server Count
+* Number of Classes
+* Number of Used Classes
+
+You can access more details on your libraries by going to the [Contrast API documentation](https://api.contrastsecurity.com/#) **> Library > Libraries section**, or using the command line. 
+
+> **Example:** To find related application names and export the data to a CSV, use a cURL request to the API: 
+```
+$ curl -H "Authorization: $(echo -n $username:$servicekey
+base64)" -H "API-Key: $apikey" https://app.contrastsecurity.com/Contrast/api/ng/4e55d28e-55e9-4b23-816b-cb4737733a47/libraries/filter?expand=apps
```
gilday: We need to replace 4e55d28e-55e9-4b23-816b-cb4737733a47 with $org_id

content/user/apps/TrackLibraries.md
terra4383: @gilday I will capture this in a ticket next week.

gilday: "The following cURL request retrieves a list of libraries where each library includes a list of applications which use that library. The jq tool formats the data as CSV for use in a custom report"

content/tools/api/AboutRestAPI.md
jday3: Should be "this version executes API requests against a real server."

PR #329, opened 2017-10-30
CONTRAST-18568 had a request for more clarification on configuration

palmergs:
I had a request to add additional clarification on how the service loads the configuration file.


kaitlynmortimer:
@palmergs Is "ruby-service" (lowercase, hyphenated) the standard way to refer to this? If so, I'll update the subhead and keep this in mind for future docs.

palmergs:
@kaitlynmortimer good catch. The gem "contrast-service.gem" should be hyphenated by not a generic reference to the "ruby service".

PR #355, opened 2017-12-28
Updates per feedback on support ticket #7526

scorlett:
Added an additional section to make clear that the -javaagent property needs to be added to the JVM args and that the appname and servername must be set.

Removed the version from the example. Without this line, the latest version will be used and avoid confusion.

Also made a slight correction - it's the API URL which needs to be set accordingly on SaaS, not the contrast.server property.

### DIFF
--- a/content/tools/sdk/MavenPlugin.md
+++ b/content/tools/sdk/MavenPlugin.md
@@ -61,7 +61,7 @@ An example configuration for the Contrast Maven Plugin:
         <username>test_user</username>
         <apiKey>testApiKey</apiKey>
         <serviceKey>testServiceKey</serviceKey>
-        <apiUrl>http://app.contrastsecurity.com/Contrast/api</apiUrl>
+        <apiUrl>https://app.contrastsecurity.com/Contrast/api</apiUrl>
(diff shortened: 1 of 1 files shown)


content/tools/sdk/MavenPlugin.md
```diff
@@ -61,12 +60,26 @@ An example configuration for the Contrast Maven Plugin:
...
     </configuration>
 </plugin>
 ```
+
+## Use the agent
+
+The final step is to add the agent to the JVM arguments used when the application is run. You must specify the following properties:
+
+```
+-javaagent:/target/contrast.jar -Dcontrast.appname=MyAppName -Dcontrast.server=MyServerName
```
ElizabethHughes: @scorlett , do we have customers always setting appname now?

scorlett: This is just when using the Maven plugin, which checks the vulns in TeamServer and will pass/fail the build based on what's been found. You have to specify the appName of the application that you want the plugin to check on the TeamServer side. So they don't _have_ to specify the appname when you run the agent, it's just so much easier if they do as they can set the Maven plugin config to the same thing.

kaitlynmortimer:
Documenting approval from Frank, for the sake of posterity.

PR #361, opened 2018-01-16
contrast.reporting.period is 5 seconds

citizenjosh:
says Elizabeth :-)


kaitlynmortimer:
According to @ScoRobi, these files are auto-generated. He's made a java-agent change and will put that into PR.

PR #387, opened 2018-03-08
update configuration to include common config options supported

cixel:


content/installation/node/NodeConfig.md
```diff
@@ -4,71 +4,89 @@ description: "Configuring the Node.js Agent"
 tags: "installation NodeJS agent configuration"
 -->
 
-Configuration options may be used to alter Contrast's behavior. They can all be appended to your startup command (e.g: ```node-contrast server.js --mute --policy foo.json```). With the exception of ```--configFile```, they can all be added to your *contrast.json* file as well:
+Configuration options may be used to alter Contrast's behavior. They can all be appended to your startup command (e.g: `npm run contrast -- --agent.loger.stdout false`, or `node-contrast server.js --agent.logger.stdout false`). With the exception of `--configFile`, they can all be added to your *contrast.json* file as well:
```
palmergs: `agent.loger.stdout` => `agent.logger.stdout`

PR #408, opened 2018-04-18
.Net Azure Agent Documentation

ldemidov:


kaitlynmortimer:
@ldemidov If we're including images for the Express installation, should we include some for the Manual version as well? Or was that intentional?

ldemidov:
No, I don't think there's a need.  There are multiple ways to add a Nuget package and we shouldnt shoehorn into one way.  The Manual installation is basically for developers who should already know how to do this anyway.

PR #410, opened 2018-04-19
:memo: Include note for Proguard users

gilday:
Proguard users must know that Proguard optimizations can cause incompatibilities with the Contrast Java agent.

aside: my editor wraps markdown at column 80. Do we have any Markdown style guides that I should be using? We should consider using a CONTRIBUTING.md to communicate Markdown styles and consider using markdown lint to enforce them

### DIFF
--- a/content/installation/java/JavaSupportedTechnologies.md
+++ b/content/installation/java/JavaSupportedTechnologies.md
@@ -34,3 +34,11 @@ JRockit 1.5, 1.6 (update 95+ or R28.3.6+) |Jetty 6, 7, 8, 9 | Struts, Struts 2
 Contrast tests WebSphere on Windows and Linux, and offers limited support for zSeries and AIX environments. Customers using WebSphere on SPARC Solaris require version 8.5.5.11.
 
 
+### Note for Proguard Users
+
+[Proguard](https://sourceforge.net/projects/proguard/files/) includes Java
+bytecode optimization features which break basic assumptions that runtime agents
+like Contrast rely on. Proguard users that wish to protect their applications
+with Contrast need to avoid these optimizations by using Proguard's
+[-dontoptimize](https://www.guardsquare.com/en/proguard/manual/usage#dontoptimize)
+configuration option.


content/installation/java/JavaOverview.md
```diff
@@ -8,3 +8,13 @@ The Contrast Java agent analyzes the behavior of Java applications running on J2
 
 ## How it Works
 To start analyzing an application, install the Java agent by adding it as `-javaagent` JVM parameter. After installation, continue to use the application as you normally would. (If it's a web application, just click on links, submit forms using normal data, and so forth.) The Java agent's sensors gather information about the application's security, architecture and libraries. You can see the results of the agent's analysis in the Contrast interface.
+
+## Note for Proguard Users
```
kaitlynmortimer: I'd make this a lower-level subhead (###).

content/installation/java/JavaOverview.md
```diff
@@ -8,3 +8,13 @@ The Contrast Java agent analyzes the behavior of Java applications running on J2
...
+
+[Proguard](https://sourceforge.net/projects/proguard/files/) includes Java
+bytecode optimization features which perform which break basic assumptions that
+runtime agents like Contrast rely on. Proguard users that wish to protect their
```
kaitlynmortimer: I'm guessing that "which perform which break" is a typo.

kaitlynmortimer:
@gilday We don't have a markdown style guide (at least not one that I know of), but I'm happy to make one. I also have no idea how markdown lint works, but happy to talk to you more about that, too!

content/installation/java/JavaOverview.md
gilday: Because it is semantically a child element of the "How it Works" heading, or just because we want smaller type?

kaitlynmortimer: I was reading it as a subsection of "How It Works" - maybe because of the "Note" in the subhead. 

If that doesn't make sense from a technical/content perspective, I'm good to keep it as a level two. But, we should probably rename it something like "Proguard Optimizations" or something similar.

gilday: I don't have strong feelings about whether or not it is a sub-heading of "How it Works" or not. It sort of feels like something miscellaneous that could belong anywhere 🤷‍♂️ - is there a better place to put it?

content/installation/java/JavaOverview.md
gilday: Fixed

gilday:
@kaitlynmortimer  I moved this note to the Technologies page

PR #417, opened 2018-04-27
Update to address new failure case seen in the wild

nahsra:


content/troubleshooting/java/installation/NoClassDefFoundError.md
```diff
@@ -9,9 +9,8 @@ tags: "troubleshoot java agent NoClassDefFoundError"
...
 ```
 
-This error indicates you've renamed your agent jar from its original name, ***contrast.jar***. 
-
+This error usually indicates you've renamed your agent jar from its original name, ***contrast.jar***. It may also mean that you're using an OSGi container like Felix or Equinox that is having trouble working with our auto-updating agent.
 
 ## Solution
 
-The name of this file can't be changed.
+The name of the contrast agent file can't be changed, and OSGi applications should only use the non-auto updating agent (also called the "1.5 agent".
```
gilday: Oh no naming crisis time. Can we say:

> should only use the "Java 1.5" agent which has the same functionality as the "Java" agent but does not auto-update.

I'm trying to avoid using any names that appear any differently than they do in the Add Agent wizard which uses "Java" and "Java 1.5"

content/troubleshooting/java/installation/NoClassDefFoundError.md
```diff
@@ -9,9 +9,8 @@ tags: "troubleshoot java agent NoClassDefFoundError"
 ```java.lang.NoClassDefFoundError:com/.../EventController
 ```
 
-This error indicates you've renamed your agent jar from its original name, ***contrast.jar***. 
-
+This error usually indicates you've renamed your agent jar from its original name, ***contrast.jar***. It may also mean that you're using an OSGi container like Felix or Equinox that is having trouble working with our auto-updating agent.
```
gilday: Did we test equinox?

nahsra: Yes, equinox is tested as part of WSO2 / Synapse.

gilday: Do you know why that test didn't catch this bug?

kaitlynmortimer:
Please note that all changes must undergo editorial review before they're merged. If I'm unavailable and the updates are urgent, please review them with Andria and keep the branch after merging; I'll then make any edits as quickly as possible after I return.

PR #427, opened 2018-05-15
Contrast 22312 api option

kaitlynmortimer:


content/installation/setup/auth/ConfigureAD2.md
```diff
@@ -1,136 +1,122 @@
...
 
 <a href="assets/images/KB4-c09_2.png" rel="lightbox" title="Configuring Groups"><img class="thumbnail" src="assets/images/KB4-c09_2.png"/></a>
 
-> **Note:** You can use the **Query for Groups** button to enable a live search of existing groups as you begin to type within the input fields.
+* **Contrast User Group:**: This group allows users to be added to an organization and log in to the standard user interface. Most users should be part of this group. 
+
+* **Contrast SuperAdmin Group:** This group allows users to log in to the Super Administrator interface. This interface should only be accessible to Application and System Administrators.
 
-## Advanced Settings
+> **Note:** A user can't belong to both the User and Admin groups. **If you add them to both groups in your AD instance, Contrast won't be able to find them during configuration.**
```
francisco-pgallardo: If you add it in both groups, it will continue to work, only that the user will act as part of the Superadmin Group

PR #443, opened 2018-06-08
Update JIRA.md

j0nS3idman:
Add an instruction for the 2-way JIRA integration to set the correct event type to fire on.  I set up a vanilla JIRA instance and the webhook didn't fire without this option selected.


gmoore:
Thanks @j0nS3idman

PR #445, opened 2018-06-11
updated environment instructions/descriptions for agents

kaitlynmortimer:
https://contrast.atlassian.net/browse/CONTRAST-23446


kaitlynmortimer:
@DanFiedler Would you prefer to include the instructions for Java in the table, instead of the paragraph above it, as suggested in https://github.com/Contrast-Security-OSS/docs/pull/447/files?

PR #450, opened 2018-06-15
Fix contrast.appname

sourabhkatti:
The -Dcontrast.appname property is deprecated, we should change them to -Dcontrast.standalone.appname or -Dcontrast.override.appname, whichever is correct for the situation. I guessed which property would work but someone with more experience with these frameworks should verify my suggestions are correct.

### DIFF
--- a/content/installation/java/installation/DesktopStandaloneApp.md
+++ b/content/installation/java/installation/DesktopStandaloneApp.md
@@ -4,9 +4,9 @@ description: "Instructions for running Java agent on desktop or standalone app"
 tags: "java agent desktop installation"
 -->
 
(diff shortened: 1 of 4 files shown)


evamill:
@sourabhkatti Is this ready to go? @gilday is a reviewer. I am, too. Looks like no one signed off yet, but we'd like to clean up this PR, if possible. Copying @terra4383

gilday:
We're inconsistently using "standalone" and "override", and we're using the legacy property names instead of the new names. We should always use the new, standalone property. As a system property, it is `-Dcontrast.agent.java.standalone_app_name`

PR #453, opened 2018-06-20
Update intructions to reference yaml instead of json.

Contrast-AlexB:
Updated instructions for bluemix to reference yaml file instead of json. Saw this on a call today.


kaitlynmortimer:
Looks good to me. Feel free to merge whenever you and @cixel think it's ready to go.

PR #477, opened 2018-07-30
Contrast 24618 release notes 354

kaitlynmortimer:


DanFiedler:
Looks good but I wonder if there's another way to write this sentence for the Java agent:
"The Java team reduced the amount of memory used by agent’s Assess analysis, and fixed an issue where certain java.lang.String methods could cause application errors, if those methods were added as sanitizers or validators."

The middle part "java.lang.String methods could cause application errors" is alarming until you read the trailing 'if' clause. I'm worried readers may not catch this important classification because of the length of the sentence.

PR #481, opened 2018-08-01
Contrast 23187 route coverage

kaitlynmortimer:
PR for page/navigation work: https://bitbucket.org/contrastsecurity/open-docs/pull-requests/164


gilday:
@kaitlynmortimer I feel we need a section in the docs which speaks to compatibility. For example, @ScoRobi is adding Route Coverage support for Spring 4, and users should not expect other frameworks to work.

kaitlynmortimer:
@ScoRobi @jleo32 @donniepropst @ldemidov 
Please advise on any updates since I opened this PR (quite a while ago). And, if you notice that I've forgotten anyone, please add them!

content/user/apps/Route-coverage.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Agent configuration 
+
+Contrast supports route coverage for the following frameworks: 
+
+* Java Spring Spring MVC 4.x 
+* .NET MVC (versions 4 and 5), WebForms, Web API and WCF
+* Node Express 
+* Ruby Rails and Sinatra
+* Python, Django, Pyramid and Flask
```
jleo32: I think this should be something like Python: Django, Flask, and Pyramid.

Python is the language and the other three are frameworks.

content/user/apps/Route-coverage.md
```diff
@@ -0,0 +1,40 @@
+<!--
+title: "Route Coverage"
+description: "Overview of setting up environments"
+tags: "user UI applications route coverage exercised vulnerabilities"
+-->
+
+## About Route Coverage
+
+An application “route” is a combination of three distinct data points: the normalized URL of the route, the HTTP verb associated with the request (e.g., Get or Post) and the method signature of the controller action. When you consistently exercising each route in your application, the Contrast agent can successfully Assess and Protect the surface layer of your application, and discover vulnerabilities.
```
ScoRobi: The route URLs may not be normalized. So we should change "normalized URL" -> "URL".

ScoRobi: We sometimes will provide data that is not the "method signature" for the signature portion of the route. This could be the location and name of an XML file or some other identifying information specific to that route, but not always a "method signature".

I would suggest to re-word to something closer to: "... and a unique signature based on that route's controller action"

content/user/apps/Route-coverage.md
```diff
@@ -0,0 +1,40 @@
...
+
+Contrast supports route coverage for the following frameworks: 
+
+* Java Spring Spring MVC 4.x 
+* .NET MVC (versions 4 and 5), WebForms, Web API and WCF
+* Node Express 
+* Ruby Rails and Sinatra
+* Python, Django, Pyramid and Flask
+
+Coverage is enabled automatically for .NET or Node applications. For Java applications, you must use the following command to specify the application name when deploying the Java agent: `-Dcontrast.appname = <example name>`.
```
ScoRobi: Coverage is now automatically enabled for Java applications although it will only properly report routes with Spring MVC 4.x applications. However coverage can be disabled with: `-Dcontrast.agent.route_coverage.enable=false`

content/user/apps/Route-coverage.md
```diff
@@ -0,0 +1,40 @@
...
+
+Contrast supports route coverage for the following frameworks: 
+
+* Java: Spring Spring MVC 4.x 
+* .NET: ASP.NET MVC (versions 4 and 5), WebForms, WebAPI and WCF
+* Node: Express 
+* Ruby: Rails and Sinatra
+* Python: Django, Pyramid and Flask
+
+Coverage is enabled automatically for .NET or Node applications. For Java applications, you must use the following command to specify the application name when deploying the Java agent: `-Dcontrast.appname = <example name>`.
```
hmcginnis: Coverage is automatic for Ruby & Python as well. Do we only want to mention the Java part since it deviates from the norm?

PR #486, opened 2018-08-08
CONTRAST-25851 the configuration option is more strict than originall…

palmergs:
Clarifying the behavior if this configuration is read by the service.


kaitlynmortimer:
@palmergs Are we ready to merge?

PR #491, opened 2018-08-15
:memo: CONTRAST-26111 Verify Java agent Exec Helper installation

gilday:
Adding instructions for the user to verify their installation of the Contrast Java agent Exec Helper. Additionally, added a note to explain that Contrast Java agent Exec Helper is only enabled in new shells

In developing the command for users to verify the Contrast Java agent Exec Helper, I made the assumption that the user is likely to copy paste this command. Given this assumption, I felt free to make this command a bit more complicated than it needs to be in order to improve the output it generates.

The command I included is `bash -c "java -Dcontrast.stdout=true -version 2>1 | grep Contrast | head -n 1"` which in addition to being more complicated, assumes that the user has programs `grep` and `head` available (which is a safe assumption, but not a guarantee). This command generates one neat line of output

> [Contrast] Wed Aug 15 17:37:23 UTC 2018 No TeamServer configuration detected. Agent will only be reporting to local listeners (e.g., Eclipse Plugin).

Alternatively, the simplest command that could work is `bash -c "java -Dcontrast.stdout=true -version"`, but it vomits a huge number of console messages which I feel distracts the user from the task at hand

### DIFF
--- a/content/installation/java/Java-linux-packages.md
+++ b/content/installation/java/Java-linux-packages.md
@@ -65,6 +65,15 @@ echo "deb https://contrastsecurity.jfrog.io/contrastsecurity/debian-staging/ $(b
 sudo apt-get update && sudo apt-get install contrast-java-agent-exec-helper
 ```
 
+* Verify that the Contrast Java Exec Helper is working by executing `java` and confirming that Contrast starts by observing console messages
+
+```
+$ bash -c "java -Dcontrast.stdout=true -version 2>1 | grep Contrast | head -n 1"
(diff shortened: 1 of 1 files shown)


kaitlynmortimer:
@gilday Do we still need to mention the where to find the logs? If so, I can link to the troubleshooting doc there. If not, I'll find another place.

gilday:
🤦‍♂️ I forgot that part. I'll send another commit

gilday:
@kaitlynmortimer I added notes about checking the logs

PR #511, opened 2018-09-18
Contrast 26040 common config

kaitlynmortimer:
To do: 
- Update file location and any other usage info
- Verify order of precedence for each agent
- Verify any added/updated properties in "Properties" and "Template" articles


content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,204 @@
...
+
+1. Corporate rule (e.g., expired license overrides `assess.enable`)
+1. System property value
+1. Environment variable value
+1. User configuration file value 
+<!-- TODO , for the java agent this includes the legacy file and the YAMl file. Should we specify both or just not mention the legacy file (which no one should be using anymore, I think.-->
+1. Contrast UI value
+1. Default value
+
+<!-- Other use and file location info. This will be done after CONTRAST-26125  -->
```
kaitlynmortimer: @ElizabethHughes Is this info available now?

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,204 @@
...
+
+Use the properties in this section to control the way and frequency with which the Java agent communicates to logs and to the Contrast UI.
+If these values are not set, the agent will use the values set in the Contrast UI.
+
+#### Diagnostic logging
+
+Use the properties in this section to control diagnostic logging. These logs allow us to diagnose any issues you may be having with the agent.
+
+<!-- Maybe a question for Jenny, should we include the 'agent' node here? -->
+
```
kaitlynmortimer: @jennymahmoudi Thoughts on this comment? And, if we should keep the top-level "agent" bullet here, should we add it to the lists in the subsections as well?

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,204 @@
...
+## Load Path
+
+Configuration values use the following order of precedence: 
+
+1. Corporate rule (e.g., expired license overrides `assess.enable`)
+1. System property value
+1. Environment variable value
+1. User configuration file value 
+<!-- TODO , for the java agent this includes the legacy file and the YAMl file. Should we specify both or just not mention the legacy file (which no one should be using anymore, I think.-->
```
kaitlynmortimer: @ElizabethHughes or @DanFiedler Can we resolve this comment?

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,204 @@
...
+
+Go to the [Java YAML Template](installation-netconfig.html#net-template) for fully formatted properties that you can copy and use in your own agent configuration files. 
+
+## Set as System Properties
+
+You can also set all of the following YAML properties as system properties. Derive the system property key from the YAML by joining every node with a "." until you reach the bottom property. 
+
+> **Example:** If you want to override the `contrast` property, as given below, you can set `-Dcontrast.enable=false` as a system property.
+  * contrast: 
+    * enable: `true`
```
kaitlynmortimer: @palmergs Is the info in this "System Properties" section true for Ruby and Python as well?

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,204 @@
...
+      * **port**: Set the port of the Syslog server to which the agent should send messages.
+      * **facility**: Set the facility code of the messages the agent sends to Syslog.
+      * **severity_exploited**: Set the log level of Exploited attacks. Value options are `ALERT`, `CRITICAL`, `ERROR`, `WARNING`, `NOTICE`, `INFO`, and `DEBUG`.
+      * **severity_blocked**: Set the log level of Blocked attacks. Value options are `ALERT`, `CRITICAL`, `ERROR`, `WARNING`, `NOTICE`, `INFO`, and `DEBUG`.
+      * **severity_probed**: Set the log level of Probed attacks. Value options are `ALERT`, `CRITICAL`, `ERROR`, `WARNING`, `NOTICE`, `INFO`, and `DEBUG`.
+
+
+<!-- ### Agent-Specific Properties
+
+That work will be done in CONTRAST-26166
```
kaitlynmortimer: @ElizabethHughes Is this done?

content/installation/net/configuration/YAML-properties-net.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Set as System Properties
+
+You can also set all of the following YAML properties as system properties. Derive the system property key from the YAML by joining every node with a "." until you reach the bottom property. 
+
+> **Example:** If you want to override the `contrast` property, as given below, you can set `-Dcontrast.enable=false` as a system property.
+  * contrast: 
+    * enable: `true`
+
+<!-- The precedence and system properties example are copied from Java. Edit or delete as needed.  -->
```
kaitlynmortimer: @DanFiedler If the precedence and system properties info is good, I'll delete this comment.

content/installation/node/YAML-properties-node.md
```diff
@@ -0,0 +1,162 @@
+<!--
+title: "YAML configuration properties for the Node agent "
+description: "Instructions for configuring Node agent properties via YAML file"
+tags: "installation node agent YAML configuration rules properties"
+-->
+
+Contrast support YAML-based configuration for the Node agent. This allows you to store configuration on disk with environment variable and CLI args. overrides.
+
```
kaitlynmortimer: Is there a better way to say "environment variable and CLI args. overrides."?

content/installation/node/YAML-properties-node.md
```diff
@@ -0,0 +1,162 @@
...
+## Set as System Properties
+
+You can also set all of the following YAML properties as system properties. Derive the system property key from the YAML by joining every node with a "." until you reach the bottom property. 
+
+> **Example:** If you want to override the `contrast` property, as given below, you can set `-Dcontrast.enable=false` as a system property.
+  * contrast: 
+    * enable: `true`
+
+<!-- The precedence and system properties example are copied from Java. Edit or delete as needed.  -->
+
```
kaitlynmortimer: @meandthemoon Can you confirm that this is correct?

content/installation/net/configuration/YAML-properties-net.md
DanFiedler: @kaitlynmortimer .NET doesn't support command line arguments. (which is fine because Windows/IIS users don't typically use command line arguments.)

content/installation/java/configuration/YAML-properties-java.md
jennymahmoudi: I thought making a change like that would require updates to all the agents, which is out of scope right now. What's the reason for wanting to remove the agent section header?

content/installation/node/YAML-properties-node.md
jennymahmoudi: How about "This allows you to store configuration on disk that you can override with environment variables or command line arguments."

Also, should that be "Contrast supports"?

content/installation/java/configuration/YAML-properties-java.md
kaitlynmortimer: @ElizabethHughes ?

content/installation/java/configuration/YAML-properties-java.md
ElizabethHughes: Let's not mention the legacy file and have line 18 be "YAML configuration file value" or something similar

content/installation/node/YAML-properties-node.md
meandthemoon: Yep - the system properties bit was missed. I'll delete it.

content/installation/java/configuration/YAML-properties-java.md
ElizabethHughes: The agent is already the top node for the logger and the security_logger. The question is just how do we tell the user that they can't just put 
```
security_logger:
     path:
```
They need to put the `security_logger` under the `agent` node.

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,203 @@
...
+Configuration values use the following order of precedence: 
+
+1. Corporate rule (e.g., expired license overrides `assess.enable`)
+1. System property value
+1. Environment variable value
+1. YAML configuration file value
+1. Contrast UI value
+1. Default value
+
+<!-- Other use and file location info. This will be done after CONTRAST-26125  -->
```
ElizabethHughes: An update here:

> The path to the YAML configuration file can be set using the environment variable `CONTRAST_CONFIG_PATH` or the java system property `contrast.config.path`. Like the rest of the configuration values, the system property takes precedences over the environment variable if both are set.

> If the environment variable and the system property aren't set, the agent will look for the YAML configuration file in the default location. 
> On Windows, the path for this is

    %ProgramData%\Contrast\java\contrast_security.yaml

> On Unix/Linux file systems, the path for this is

    /etc/contrast/java/contrast_security.yaml

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,203 @@
...
+
+Go to the [Java YAML Template](installation-netconfig.html#net-template) for fully formatted properties that you can copy and use in your own agent configuration files.
```
ElizabethHughes: @kaitlynmortimer , I'm not sure how to comment on this template, but can you update it with the latest from my last PR, please? I'll send you the file privately. Thanks!

ElizabethHughes: The other agents should also make sure these Docs have the latest templates.

content/installation/java/configuration/YAML-properties-java.md
ElizabethHughes: I don't think it is. These system properties are a java thing, and I'm not sure what the ruby equivalent is.

content/installation/java/configuration/YAML-properties-java.md
ElizabethHughes: Yep. 
```
  #==========================================================================
  # Java
  # The following properties apply to any Java agent-wide configurations.
  #==========================================================================
  # java:

    # Set the name of a standalone application. If this value is set, the agent assumes there is only one application in this server.
    # standalone_app_name: NEEDS_TO_BE_SET

```

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,203 @@
...
+    * **user**: Set the proxy user.
+    * **pass**: Set the proxy password.
+    * **auth_type**: Set the proxy authentication type. Value options are `NTLM`, `Digest`, and `Basic`.
+
+
+### Contrast agent properties
+
+Use the properties in this section to control the way and frequency with which the Java agent communicates to logs and to the Contrast UI.
+If these values are not set, the agent will use the values set in the Contrast UI.
+
```
ElizabethHughes: We've added two properties under the `agent` node. Should we add those here? The properties are
```
agent:

  # Set how long to run the agent before shutting down itself (in milliseconds). A negative value disables scheduled shutdown.
  shutdown_time_ms: NEEDS_TO_BE_SET

  # Enable to deinstrumentation classes on shutdown. If this is not enabled, the agent disables sensors on shutdown, but leaves instrumentation.
  deinstrument_on_shutdown: NEEDS_TO_BE_SET
```

ElizabethHughes: We'll need to update the description on line 63 too. Maybe something like

> Use the properties in this section to control agent wide behaviors. For instance, this section can be used to control the way in which the agent starts up and shuts down, and the way in which it communicates to logs and to the Contrast UI.

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,203 @@
...
+Use the properties in this section to control Assess in the Java agent. The sampling settings allow you to control which requests the agent tracks and which it ignores. The rules setting allows you to control which Assess rules are disabled. 
+
+> **Note:** If you need a complete list of rules, use the **Support** widget in OpenDocs to contact Contrast's Customer Support team.  
+
+
+* **assess**:
+
+  * **enable**: Include this property to determine if the Assess feature should be enabled. If this property is not present, the decision is delegated to the Contrast UI. <br> Example: `true`
+  * **tags**: Apply a list of labels to vulnerabilities and preflight messages. Labels must be formatted as a comma-delimited list. Example: `label1, label2, label3`
+  * **stacktraces**: Value options are `ALL`, `SOME`, or `NONE`.
```
ElizabethHughes: This line needs to be removed. The java agent doesn't support `assess.stacktraces`.

content/installation/java/configuration/YAML-properties-java.md
```diff
@@ -0,0 +1,203 @@
...
+
+    * **method-tampering**:
+      * **mode**: Set the mode of the rule. Value options are `monitor`, `block`, `block_at_perimeter`, or `off`. <br> Note: If a setting says, "if blocking is enabled", the setting can be `block` or `block_at_perimeter`.
+
+    * **reflected-xss**:
+      * **mode**: Set the mode of the rule. Value options are `monitor`, `block`, `block_at_perimeter`, or `off`. <br> Note: If a setting says, "if blocking is enabled", the setting can be `block` or `block_at_perimeter`.
+
+    * **xxe**:
+      * **mode**: Set the mode of the rule. Value options are `monitor`, `block`, `block_at_perimeter`, or `off`. <br> Note: If a setting says, "if blocking is enabled", the setting can be `block` or `block_at_perimeter`.
+
```
ElizabethHughes: Can you add
```
    #=============================================================================================
    # Csrf
    # Use the following properties to configure how the cross-site request forgery rule works.
    #=============================================================================================
    csrf:

      # Set the mode of the rule. Value options are `monitor`, `block`, `block_at_perimeter`, or `off`. 
      #  Note - If a setting says, "if blocking is enabled", the setting can be `block` or `block_at_perimeter`.
      mode: monitor
```

as another rule, please.

content/installation/net/configuration/YAML-template-net.md
```diff
@@ -0,0 +1,359 @@
...
+
+
+#================================================================================
+# Contrast
+# Use the properties in this section to connect the agent to the Contrast UI.
+#================================================================================
+contrast:
+
+# Only set this property if you want to turn off Contrast. Set to `true` to turn the agent on; set to `false` to turn the agent off.
+# enable: true
```
DanFiedler: It looks like everything in this template file is left aligned. This breaks the yaml format and parsing in the agent. 

The alignment should be unchanged from what is produced by the spec parser.

For example, I'd expect:
`
contrast:
  enable: true
`

Rather than:
`
contrast:
enable: true
`

DanFiedler:
It looks like everything in the .NET template file is left aligned. This breaks the yaml format and parsing in the agent.

The same formatting problem appears to be present in the other templates too.

Can't approve until we resolve this.

kaitlynmortimer:
@DanFiedler The issue should be resolved. Please verify.

DanFiedler:
> @DanFiedler The issue should be resolved. Please verify.

Templates look good.

PR #517, opened 2018-09-24
coverage note re: unsupported frameworks

kaitlynmortimer:
Article at https://docs.contrastsecurity.com/user-apps.html#route states, "For unsupported frameworks, Contrast displays the observed routes only." Statement must be revised to reflect more accurate explanations given by each agent/engineer.


ldemidov:
I would just revise that whole sentence and add more detail.  Instead, say something like this:

For supported frameworks Contrast can discover all possible routes in your application at startup.  To build the coverage report Contrast will then observe these routes being exercised as the application runs.  

For unsupported frameworks, Contrast cannot discover all possible routes so we cannot determine which routes have not been exercised.  When the application runs, we will estimate the observed route by using application stack traces.

PR #485, opened 2018-08-08
Debian and RPM Package Installation for Contrast Java agent

kaitlynmortimer:
Page/navigation work: https://bitbucket.org/contrastsecurity/open-docs/pull-requests/171/contrast-22950-linux-pkgs/diff

PR #524, opened 2018-10-05
webserver agent install documentation

seschis:
Adds documentation for how to install, configure, and run the webserver agent.  Focuses on install from linux packages.

Related Tickets:
CONTRAST-27468

### DIFF
--- a/content/installation/linux_package_repo/ContrastRepoConfig.md
+++ b/content/installation/linux_package_repo/ContrastRepoConfig.md
@@ -1,0 +1,51 @@
+Some of Contrast's products are distributed via Linux packages. This makes
+installing and staying up to date very easy. In order to use our package repo
+you have to configure your package management system to include a reference to
+our package repo. The instructions for doing this are below:
+
+Some of our packages are tied to the specific Linux distribution you are using;
+For Ubuntu, this could be one of 'bionic', 'xenial', or 'trusty'.  The
+instructions will show you how to get this information and then use it to
+configure the right package repo for your system.  This is a one-time
+configuration.  Once its done, you'll be connected to the software distribution
+pipeline and can solely manage it via your distribution's package manager.
+
+### Debian/Ubuntu
+
+
+* Configure your system to retrieve from the correct Debian repository. Get the `CODENAME` for your Ubuntu/Debian release. 
+
+```
+grep VERSION_CODENAME /etc/os-release 
+```
+
+* Update the command below with the `CODENAME`, and run the commands.
+
+```
+curl https://contrastsecurity.jfrog.io/contrastsecurity/api/gpg/key/public | sudo apt-key add -
+echo "deb https://contrastsecurity.jfrog.io/contrastsecurity/debian-public/ CODENAME contrast" | sudo tee /etc/apt/sources.list.d/contrast.list
+```
+
+* Once you've finished configuration, update your local package cache
+
+```
+sudo apt-get update
+```
+
+### RedHat/CentOS
+
+Complete the following steps to install the Contrast Service for Red Hat Enterprise Linux (RHEL) and CentOS versions 5, 6 and 7.
+
+* To install Contrast Service from Contrast's Yum repository, configure your system to use the repository.
+
+```
+OSREL=$(rpmquery -E "%{rhel}")
+sudo -E tee /etc/yum.repos.d/contrast.repo << EOF
+[contrast]
+name=contrast repo
+baseurl=https://contrastsecurity.jfrog.io/contrastsecurity/rpm-public/centos-$OSREL/
+gpgcheck=0
+enabled=1
+EOF
+```
+
--- a/content/installation/linux_package_repo/ContrastRepoConfig.md
+++ b/content/installation/linux_package_repo/ContrastRepoConfig.md
@@ -4,7 +4,7 @@ you have to configure your package management system to include a reference to
 our package repo. The instructions for doing this are below:
 
 Some of our packages are tied to the specific Linux distribution you are using;
-For Ubuntu, this could be one of 'bionic', 'xenial', or 'trusty'.  The
+For example, on Ubuntu this could be one of 'bionic', 'xenial', or 'trusty'.  The
 instructions will show you how to get this information and then use it to
 configure the right package repo for your system.  This is a one-time
 configuration.  Once its done, you'll be connected to the software distribution
--- a/content/installation/webserver-agent/WebserverAgentConfig.md
+++ b/content/installation/webserver-agent/WebserverAgentConfig.md
@@ -1,0 +1,105 @@
+# Configuration
+
+The items that must be configured are:
+* The communication link between the Webserver Agent and Contrast-Service
+* The communication link between Contrast-Service and TeamServer
+* The NGINX service enabling the Agent to inspect traffic to certain endpoints
+
+# Configuring Contrast-Service
+
+_XXX: config location is moving to
+/etc/contrast/webserver/contrast-security.yaml_
+
+Contrast-Service is controlled by the configuration located at
+`/etc/contrast/contrast_security.yaml`. This is the Contrast Common Config
+format that most agents utilize.  _XXX: link to a general common config
+document_
+
+This file will control how the
+Webserver is represented to TeamServer. The default configuration installed
+with the contrast-service linux package has most necessary items filled in
+however you will need to edit it with your TeamServer location and API key. You
+will also need to configure how you want your agent represented to TeamServer
+
+    server:
+      name: YOUR SERVER NAME
+      path: /
+      type: Proxy
+      environment: development
+
+The other section that needs your specific config in this file is:
+
+    contrast:
+      user_name: YOUR_CONTRAST_USERNAME
+      service_key: YOUR_CONTRAST_SERVICE_KEY
+      api_key: YOUR_CONTRAST_API_KEY
+      url: http://YOUR_TEAMSERVER_LOCATION/Contrast
+
+_XXX: where do I point someone if they are confused as to where to get these
+item?_
+_XXX: give the user a way to test their configuration is working wrt to
+TeamServer?_
+
+# Configuring NGINX with the Webserver Agent.
+
+The Webserver Agent is configured within the NGINX configuration files. These
+are located at `/etc/nginx`.  Configuring NGINX as a reverse proxy or webserver
+is out of scope for this document. Here we'll explain the Webserver Agent
+specific configuration within the NGINX configuration files.
+
+The /etc/nginx.conf should have the webserver agent module added to it, and
+configuration items enabling the webserver agent for certain endpoints in the
+nginx.conf.
+
+
+Example /etc/nginx/nginx.conf:
+    
+    load_module modules/ngx_http_contrast_connector_module.so;
+
+    events {
+      worker_connections 1024;
+    }
+
+    http {
+
+      contrast on;
+      contrast_debug on;
+      contrast_unix_socket "/run/contrast-service.sock";
+
+      error_log logs/error.log debug;
+
+      server {
+        listen 8888;
+        server_name localhost;
+
+        location / {
+          autoindex on;
+          index index.html index.html;
+          contrast_app_name "APP NAME STATIC ENDPOINT";
+        }
+
+        location /app {
+          rewrite /MY_APP/(.*) /$1 break;
+          proxy_set_header Host $host;
+          proxy_set_header X-Real-IP $remote_addr;
+          proxy_pass http://127.0.0.1:4567;
+          contrast_app_name "APP NAME DYNAMIC ENDPOINT";
+        }
+      }
+    }
+
+This important parts here are the `load_module` directive at the top which is
+loading the webserver agent into nginx and the various `contrast_*` directives.
+The `contrast_*` directives can be placed at the "main", "server", or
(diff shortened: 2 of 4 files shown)


content/installation/linux_package_repo/ContrastRepoConfig.md
```diff
@@ -0,0 +1,43 @@
...
+
+Some of our packages are tied to the specific Linux distribution you are using. For example, on Ubuntu this could be one of 'bionic', 'xenial', or 'trusty'.  The instructions will show you how to get this information and then use it to configure the right package repo for your system.  This is a one-time configuration.  Once its done, you'll be connected to the software distribution pipeline and can solely manage it via your distribution's package manager.
+
+### Debian/Ubuntu
+
+
+* Configure your system to retrieve from the correct Debian repository. Get the `CODENAME` for your Ubuntu/Debian release. 
+
+```
+grep VERSION_CODENAME /etc/os-release
```
palmergs: My brain is failing for a fool-proof solution, but could a carefully crafted grep be used to set an ENV variable that could be used in the following script?  e.g.

    export `grep VERSION_CODENAME /etc/os-release`
    curl https://contrastsecurity.jfrog.io/contrastsecurity/api/gpg/key/public | sudo apt-key add -
echo "deb https://contrastsecurity.jfrog.io/contrastsecurity/debian-public/ $VERSION_CODENAME contrast" | sudo tee /etc/apt/sources.list.d/contrast.list

gilday:
@kaitlynmortimer is there something in particular you want me to look at in these changes?

gilday:
@kaitlynmortimer filled me in that the intention is to consolidate our linux package documentation. @seschis @palmergs I assume this includes the yet-to-be-released Java agent packages? Documentation changes for Java linux packages are in #485. 

There is some overlap between the instruction here and the instructions in that PR such as adding the GPG key for the debian repo

    curl https://contrastsecurity.jfrog.io/contrastsecurity/api/gpg/key/public | sudo apt-key add -

But there are some differences too. For example, these instructions would have the user add a debian repo specific to the Debian release e.g. Ubuntu Xenial, but some of the Java agent packages are not in distribution specific repositories, but rather are in our distribution agnostic "all" repository.

I'm not convinced that the user's experience will be improved if we consolidate all of the package repository configuration to one place. Do you think we better serve our users if we leave the Java agent system package configuration as is vs consolidating it here?

content/installation/linux_package_repo/ContrastRepoConfig.md
seschis: I really dislike the whole pkg-repo setup thing.  Its just so 'in-the-weeds' feeling.  I'm considering just making a package that a user downloads directly and sets up the repo for them. Something similar to the epel-release package described here: https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F

content/installation/webserver/WebserverAgentConfig.md
```diff
@@ -0,0 +1,85 @@
+
+
+The items that must be configured are:
+* The communication link between Contrast-Service and TeamServer
+* The NGINX service enabling the Agent to inspect traffic to certain endpoints
+
+## Configure the Contrast Service
```
kaitlynmortimer: @palmergs Is this different from the configuration instructions at https://docs.contrastsecurity.com/installation-service.html#service-config? If so, we should explain the difference and/or how they work together.

content/installation/webserver/WebserverAgentConfig.md
```diff
@@ -0,0 +1,85 @@
...
+The other section that needs your specific config in this file is:
+
+    contrast:
+      user_name: YOUR_CONTRAST_USERNAME
+      service_key: YOUR_CONTRAST_SERVICE_KEY
+      api_key: YOUR_CONTRAST_API_KEY
+      url: http://YOUR_TEAMSERVER_LOCATION/Contrast
+
+_XXX: where do I point someone if they are confused as to where to get these item?_
+_XXX: give the user a way to test their configuration is working wrt to TeamServer?_
```
kaitlynmortimer: @palmergs Any suggestions?

content/installation/webserver/WebserverAgentRunning.md
```diff
@@ -0,0 +1,18 @@
+# Starting the services
```
kaitlynmortimer: @seschis Is this similar to any articles in the other agent sections? (I want to be sure I keep with our naming conventions.)

kaitlynmortimer:
To follow up on @gilday's concerns about moving the Java agent system package configuration instructions - 

I'm wondering if it will cause confusion to have a shared linux packages article, if it's not completely the same for each bit of Contrast, and we then have very similar info else where (i.e., the Java installation article). I could still put a blurb in https://docs.contrastsecurity.com/installation-setup.html#onboard about the non-UI option.   

@palmergs and @seschis - any thoughts?

content/installation/webserver/WebserverAgentConfig.md
palmergs: 🚧  It is almost the same. There's one additional configuration that is required for the Webserver agent.  `agent.service.socket = /run/contrast.sock` is required. Let me look at a separate PR to get that updated on that page.

seschis: Differences that I see are the location of the contrast_security.yaml file and the fact that it configures a tcp port rather than a unix socket for the agent to speak to.
Personally, I would rather the all the webserver agent configuration be centrally located and described.  At this point, as a user it feels confusing to me if I'm trying to configure the webserver agent but I need to deal with settings potentially applicable to ruby and python agents.
@palmergs we also haven't even talked about the separate modsecurity configuration files.  Do I mention those at all?  I don't really want a user poking around them, but there is nothing to stop them.  What are you're thoughts about all this?

content/installation/webserver/WebserverAgentConfig.md
seschis: For the latter item, I can tell the user to check the logs.  Wouldn't it be cool if the TS UI just sorta told us that the agent was communicating fine.  Maybe this is a fault of the agent in that it needs to send an initial 'registration' or 'ping' msg on startup.

In the short term I'll just highlight the log file to the user and tell them to check it if there are errors.

seschis:
> To follow up on @gilday's concerns about moving the Java agent system package configuration instructions -
> 
> I'm wondering if it will cause confusion to have a shared linux packages article, if it's not completely the same for each bit of Contrast, and we then have very similar info else where (i.e., the Java installation article).
> 
> In any case, I could still put a blurb in https://docs.contrastsecurity.com/installation-setup.html#onboard about the non-UI option.
> 
> @palmergs and @seschis - any thoughts?

I realize that most of our agent teams work in a silo type of fashion but I'd really like to change that with respect to our package distribution system.  I'm going to advocate to put the user thru the pain of configuring our package repo on their system only once.  Its going to avoid confusion later and allow contrast to represent a single point for the distribution of its package and how a user configures for it.

@gilday I think we can live with allowing the java team to use the 'all' repo while other products can use a '<distro>' repo depending on what makes sense for them.  There are just two lines in the /etc/apt/sources.d/contrast.list rather than one.  Configuring a linux repo just looks ugly to me and I'd rather a user do it only once and then allow them access to whatever they need/want from us.  Does this still work for the java team?
What are your thoughts on configuring a user's system for our repo via an actual package file?  This is the pattern used by the EPEL (https://fedoraproject.org/wiki/EPEL#Quickstart). Do you like this pattern? Seems a more fool-proof way then asking a user to edit the files directly and also gives us the ability to add new repo locations to their system should we ever need to when they upgrade.

content/installation/webserver/WebserverAgentConfig.md
seschis: discussed this with @palmergs.  We decided to treat the existing contrast-service doc as a "global" doc that discuss contrast-service configuration with respect to any agent.  The content for the webserver agent will stay as it represents the minimum about of info a user needs to configure the contrast-service in the role of a webserver agent.  It would easier and less confusing new users than directing them to a global/generic doc of contrast-service overall config possibilities.

content/installation/linux_package_repo/ContrastRepoConfig.md
```diff
@@ -0,0 +1,49 @@
...
+	* Ubuntu 16.04: xenial
+	* Ubuntu 14.04: trusty
+
+* First install the public key we use to sign our packages.
+
+```
+curl https://contrastsecurity.jfrog.io/contrastsecurity/api/gpg/key/public | sudo apt-key add -
+```
+
+* Next, add the package repos. Be sure to replace the `CODENAME` below with your distros actual codename. This can be found in ```/etc/os-release```. Create the file ```/apt/sources.list.d/contrast.list``` with your favorite editor and add the content below to it:
```
gilday: Instead of asking users to figure out their `CODENAME` and their favorite editor, I think we should give them a script to do this. The Java linux packages docs use

```bash
echo "deb https://contrastsecurity.jfrog.io/contrastsecurity/debian-public/ $(bash -c '. /etc/lsb-release && echo $DISTRIB_CODENAME') contrast" | sudo tee /etc/apt/sources.list.d/contrast-public-$(bash -c '. /etc/lsb-release && echo $DISTRIB_CODENAME').list
```

gilday: Should use a single back tick for inline code samples - I'm not sure if the fences style triple back ticks works inline

content/installation/linux_package_repo/ContrastRepoConfig.md
```diff
@@ -0,0 +1,49 @@
...
+
+```
```
gilday: Do we get any code highlighting benefits if we add a language directive to this code block i.e. "bash"?

\```bash
\```

I guess this is a two part question:

1. Does it make sense to highlight these commands as "bash"
1. Is our static site generator cool enough to apply code highlighting if we use these language directives?

content/installation/linux_package_repo/ContrastRepoConfig.md
```diff
@@ -0,0 +1,49 @@
...
+sudo apt-get update
+```
+
+### RedHat/CentOS
+
+Complete the following steps to install the Contrast Service for Red Hat Enterprise Linux (RHEL) and CentOS versions 5, 6 and 7.
+
+* The Contrast repositories for RHEL/Centos can be added by creating the file, ```/etc/yum.repos.d/contrast.repo```, and adding the following data to it with your favorite text editor:
```
gilday: ditto inline fences

content/installation/linux_package_repo/ContrastRepoConfig.md
gilday: I can get onboard for having all of the agents share a linux repository configuration. Is the idea that we will share this documentation by linking to it from the agent specific documentation pages?

gilday: Copy edit: we use the shorthand "repo" instead of "repository" a lot but not always. We should standardize on one, and I think we should favor the long form instead of shorthand

seschis: going to rework this section to be easier for the user but that script line looks really complex, so I'll try to simplify this some by only asking the user to know what version of Ubuntu they are running.

seschis: After refactoring from the above comment, this paragraph went away.

seschis: yes.  I think it will help unify how a user gets and installs software from contrast.  Provide a similar/familiar experience for most of our product line.

content/installation/linux_package_repo/ContrastRepoConfig.md
seschis: I was probably using the wrong MD for this.  I expected some type of highlighting to occur, like displaying it with monospaced font or grey boxing it or something.
After refactoring from above comments this stuff got nixed.

content/installation/linux_package_repo/ContrastRepoConfig.md
seschis: changed mentions of it in this doc to the long form.

content/installation/linux_package_repo/ContrastRepoConfig.md
seschis: This stuff was refactored the stuff in question was removed.

content/installation/linux/LinuxRepoConfig.md
```diff
@@ -0,0 +1,66 @@
...
+
+Some of Contrast's products are distributed through Linux packages. This makes installation and updates very easy and familiar as it utilizes the Linux packaging and distribution system. 
+
+To use the Contrast package repository, you must configure your package management system to include the appropriate reference.
+
+While Contrast packages are tied to the specific Linux distribution you're using, some are applicable to any distribution. The following steps for **Ubuntu** and **RedHat/CentOS** users configure both distribution points; once configured, you will have access to all our Linux-packaged products.
+
+**This is a one-time configuration step.** Once its done, you're connected to the Contrast software distribution pipeline, and can solely manage it through your distribution's package manager. 
+
+> **Note:** Some Linux package managers need a public GPG key that's used to verify the signed packages from our repository; this is included in the configuration steps, if its applicable to your system.
```
palmergs: it's

kaitlynmortimer: Yes. I fixed the wrong typo in my latest commit.

PR #531, opened 2018-10-17
:sparkles: Added application code configuration

ScoRobi:


gilday:
@ScoRobi should decline this change? This configuration property is already documented in our [Config YAML section](https://docs.contrastsecurity.com/installation-javaconfig.html#java-yaml), and the system properties updated in this change are deprecated and will be retried in favor of Common Config

ScoRobi:
Configuration already documented and the system properties are deprecated.

PR #547, opened 2018-11-01
add remaining certificate options

Avuxo:
Add the remaining Node config options for PKI certificates to the documentation.


kaitlynmortimer:
@Avuxo or @palmergs - Do these properties belong in the YAML file/articles as well?

PR #556, opened 2018-11-20
Remove zk docs

thedunster:
Removing documentation around this since it's not really used and not the recommended way of doing distributed teamserver.


kaitlynmortimer:
@thedunster Let's move the article to the Archive folder, if possible.

thedunster:
@kaitlynmortimer Done!

PR #565, opened 2018-12-02
Update mysql language

thedunster:


dhafley:
If you know why c++ has to get installed, it never hurts to share.  It'll also save Simon from an explanation or two.

thedunster:
I've pushed a change referencing MySQL's release notes around C++.

PR #575, opened 2018-12-11
Remove group before deleting an application

Marx314:
Not removing the group before the application make it disappear from the interface but won't let you recreate it.


kaitlynmortimer:
The issues described can't be solved by documentation updates. If the issue still exists, please send a message to bugs@contrastsecurity.com.

PR #576, opened 2018-12-13
fix: reference correct file name

MattMorgis:
The current `.tar` is named `node-contrast-1.35.2.tar`

Note the dash (`-`) instead of the underscore (`_`). 

It's also like this on `app.contrastsecurity.com`, but I couldn't find anywhere to submit a PR for this. 

Along with an instruction that doesn't work, and isn't in this `docs` repo:
>After installation, the Node.js agent can be run as follows: `node-contrast <app-main>.js`

 
screen shot 2018-12-13 at 10 08 08 am

Additionally, I get promise warnings, but again, not sure where to report or submit a fix:

```
(node:25311) Warning: a promise was created in a handler at domain.js:409:12 but was not returned from it, see http://goo.gl/rRqMUw
    at Function.Promise.cast(/node_modules/node_contrast/node_modules/bluebird/js/release/promise.js:196:13)
```

### DIFF
--- a/content/installation/node/NodeInstall.md
+++ b/content/installation/node/NodeInstall.md
@@ -9,7 +9,7 @@ tags: "NodeJS agent installation"
 After downloading from your account, install the agent from your application's root directory as follows:
 
 ``` sh
-npm install node_contrast-#.#.#.tar.gz
+npm install node-contrast-#.#.#.tar.gz
 ```
 This will add the agent to your *node_modules* folder without creating an entry in the dependencies list of your *package.json*.
 


meandthemoon:
Thanks, Matt. 

I'm actually seeing the current download as `node-contrast-1.35.2.tgz`. To my knowledge, the artifact should be named `node_contrast-1.35.2.tgz`. In any case the docs appear to need correction. I can connect with the Contrast Ops team and verify the correct name and we can go from there.

Per the `node-contrast` command, when the agent is installed a symlink is created in `node_modules/.bin/node-contrast -> ../node_contrast/cli.js`. It's not global, so it will only work when the command is invoked within the project's directory. The installer docs could be more explicit about this; I can submit a change to our internal site's content for this. Also, `<app-main>.js` is a placeholder for your app's entry point file. I don't know whether that was a point of confusion as well.

For technical support you can reach the support team at support@contrastsecurity.com. They're pretty fantastic.

MattMorgis:
Yeah, I wasn't sure if the convention in the filename was wrong or it had changed and the docs hadn't reflected it yet. It'd be best if the command and directory could be consistent, but I'm unsure if that causes any conflicts. 

I can't get the `` command to run, even within the directory, at all, unless it's an npm script. I assumed that's why that step was removed from these docs, but just maybe overlooked in the internal site.

Even from within the project directory I get:
```
node-contrast: command not found
```

MattMorgis:
FWIW, in `autoUpdate.js`, you guys download a newer version if available using the file convention:
`'node-contrast-' + available + '.tgz'`

kaitlynmortimer:
@MattMorgis I'm closing this PR as the issues discussed can be more effectively handled through our Customer Support team. You can send them an email at support@contrastsecurity.com.

cixel:
Hey @MattMorgis, you're right about the filename, and the install instructions are a holdover from when we used to recommend `npm i -g` for the agent. I'll make sure this gets fixed, both in the docs and on app.contrastsecurity.com. Thanks for the PR!

PR #577, opened 2018-12-13
added example command

kaitlynmortimer:


kaitlynmortimer:
@gilday I'm assuming more info is needed for when and how to use this example command. Any context you can add would be helpful.

content/installation/java/Java-basic-installation.md
```diff
@@ -6,16 +6,16 @@ tags: "installation Java agent overview installation"
...
-* Log in to the Contrast interface.
+* Log in to the Contrast UI.
 * Click the button in the top navigation bar to **Add Agent**.
 * Select the **Java** agent in the dropdown menu, and click the button to **Download Agent**. You may need to specify proxy authentication information required by your network before downloading the agent.
 
 <a href="assets/images/Download-java-agent.png" rel="lightbox" title="Download the Java agent"><img class="thumbnail" src="assets/images/Download-java-agent.png"/></a>
 
+* After download, add a JVM parameter to install the Java agent. 
 
-> **Note:** You must not rename the *contrast.jar* file. If the agent file name is changed, the agent will [fail to connect](troubleshooting-javainstall.html#class) with the Contrast application.
```
gilday: This change deletes an important note for our customers. Regardless of where a customer retrieves the agent jar from, it is still the case that the customer cannot rename this file. We are redesigning a core part of the agent that will soon eliminate this restriction, but for now it is a common source of support issues

content/installation/java/JavaOverview.md
```diff
@@ -6,5 +6,6 @@ tags: "installation Java agent overview introduction"
...
 
-## How it Works
-To start analyzing an application, install the Java agent by adding it as `-javaagent` JVM parameter. After installation, continue to use the application as you normally would. (If it's a web application, just click on links, submit forms using normal data, and so forth.) The Java agent's sensors gather information about the application's security, architecture and libraries. You can see the results of the agent's analysis in the Contrast interface.
+## How It Works
+
+To start analyzing an application, [install the Java agent](installation-java.html#java-standard) by adding it as `-javaagent` JVM parameter. After installation, continue to use the application as you normally would. (If it's a web application, just click on links, submit forms using normal data, and so forth.) The Java agent's sensors gather information about the application's security, architecture and libraries. You can see the results of the agent's analysis in the Contrast interface.
```
gilday: In a previous change, you replaced "Contrast interface" with "Contrast UI"

content/installation/java/Java-basic-installation.md
```diff
@@ -6,16 +6,16 @@ tags: "installation Java agent overview installation"
...
 * Click the button in the top navigation bar to **Add Agent**.
 * Select the **Java** agent in the dropdown menu, and click the button to **Download Agent**. You may need to specify proxy authentication information required by your network before downloading the agent.
 
 <a href="assets/images/Download-java-agent.png" rel="lightbox" title="Download the Java agent"><img class="thumbnail" src="assets/images/Download-java-agent.png"/></a>
 
+* After download, add a JVM parameter to install the Java agent. 
 
-> **Note:** You must not rename the *contrast.jar* file. If the agent file name is changed, the agent will [fail to connect](troubleshooting-javainstall.html#class) with the Contrast application. 
- 
+> **Example:** `java -javaagent:contrast.jar -jar <<app-name>>.jar`
```
gilday: This is a good canonical example. We have more examples tailored for specific technologies on the [Installation by Container](https://docs.contrastsecurity.com/installation-javainstall.html) page. Should we mention that page on this page?

gilday: Should we use double angle brackets i.e. `<<foo>>` to mean "replace with user value"? I believe I've used single angle bracket notation for this in other docs because I didn't know we had a convention, but I'll make sure to correct those going forward if double brackets is our convention

kaitlynmortimer: No set convention. I'll update to single.

PR #583, opened 2018-12-18
Contrast 28585 one jvm one app

phil-mcgee:
CONTRAST-28585 Java Agent should issue warnings that standalone and appname override flags will soon be deprecated. Users should move to application.name from common config

Description
The agent should:
1) detect when standalone and appname override flags are being used. 
2) if those flags are being used the agent should log to the console and agent logs to instruct users that those flags will be deprecated in an upcoming release and consolidated under application.name.

**We also need an article on docs to this move to standalone with more information for users.**

Additions to the release notes have already been submitted.  The changes in this PR are to add deprecation notes to the two properties to be (eventually) removed to the JavaAgentSystemProperties.md file and to add the new preferred property (application.name).  Each of these three items includes a TODO to add a link to the release notes (if possible).  If it is not practical to add those links the TODOs will need to be removed.


DanFiedler:
Do we have an actual link to put in the TODO spot? 

We can't really merge and publish this until there's real content in that TODO spot.

kaitlynmortimer:
@DanFiedler I removed the "TODO"s in my commit. Do you want to wait for a link to the release notes that mention this change, or just omit the links entirely?

PR #585, opened 2018-12-19
Contrast 16359 filters

kaitlynmortimer:


kaitlynmortimer:
@terra4383 The UI Intro and Navigation docs need a lot of help (outside the scope of this PR). I tried to reign myself in and refocus. I hope the changes make sense!

content/user/starthere/UI-intro.md
```diff
@@ -79,25 +77,20 @@ Contrast discovers any code flaws, which are presented with a severity level to
  
 ## Track Findings
 
-Contrast gives you the ability to send vulnerabilities to bugtracker integrations or by email for users who don't have access to Contrast. You can set up these and a bunch of other integrations - including Slack, HipChat or any generic WebHook integration - by selecting **Organization Settings** in the User menu and then **Integrations** in the sidebar. You can tell Contrast notify you if there are any new high or critical vulnerabilities found in your application. 
+Contrast gives you the ability to [send vulnerability data](user-vulns.html#track) to bugtracker integrations or to users via email. You can set up bugtrackers and other integrations - including Slack, HipChat or any generic WebHook integration - by going to the **user menu > Organization Settings > Integrations tab** in the sidebar. You can tell Contrast notify you if there are any new high or critical vulnerabilities found in your application.
```
terra4383: "Tell Contrast **to** notify you..."

content/user/starthere/UI-navigation.md
```diff
@@ -51,20 +52,20 @@ If you have system administration capabilities, you will be able to switch from
 
 ## Grid Actions
 
-Many pages within Contrast contain rows of data presented in table format. These grids have a built-in search mechanism to easily get to the data you need. You can also use the quick views to see data by the most frequently used categories. 
+Contrast displays primary assets and findings - applications, servers, libraries, vulnerabilities and attacks - in table format with rows of data. Use the search field to quickly find the data you need. You can also use the quick views to see data by the most frequently used categories.
```
terra4383: "to filter by the most frequently used categories"

content/user/starthere/UI-navigation.md
```diff
@@ -51,20 +52,20 @@ If you have system administration capabilities, you will be able to switch from
...
 
 ### Sorting and filtering 
 
-Advanced filtering is available on each main listing of assets and findings - Applications, Servers, Libraries, Vulnerabilities, and Attacks. Simply click the **Advanced** link next to the search field. Here you’ll find panels of various data sets that allow you to limit the result set and hone in on the pieces that interest you.
+Sort and filter data in the grids to adjust your focus. Use the dropdown menu above the grids to select a quick view of findings in certain categories, such as licensed applications or open vulnerabilities. You can also use the filters in the grids to find specific sets of findings, such as servers with the same tags. The dropdown menu above the right corner of the grid lets you sort all the findings in your current view by certain categories, such as application name (alphabetically), in ascending or descending order.
```
terra4383: You already referred to these as "quick views" earlier in the Grid Actions section, so here I would just say "Use the quick views at the top of the page to hone in on findings in certain categories..."  

"You can also use the column header filters to narrow down to a specific set of findings...."

"The Sort button in the top right corner of the page provides variables upon which you can choose to sort the grid data in ascending or descending order - for example, application name."

terra4383: I wonder if we should mention Select All capability per category? Maybe not, as it's probably pretty easily discoverable.

terra4383:
Ok a couple of notes but otherwise looks good! You're right that the earlier parts need help. What is the current protocol for updating a doc as a whole if it's not going to be done at the time of a change? Is it at least captured in a ticket somewhere? Thanks!

content/user/starthere/UI-navigation.md
terra4383: Another option: "You can also use the column header filter panels to locate a specific set of findings..."

kaitlynmortimer:
Yes - I'm going to open a ticket for the general improvements. Thanks!

PR #593, opened 2019-01-04
Contrast 29288 node2

Avuxo:
Documentation updates for the new agent now that we have C++ dependencies.

### DIFF
--- a/content/installation/node/NodeInstall.md
+++ b/content/installation/node/NodeInstall.md
@@ -18,11 +18,11 @@ To install the Node agent, you must have installed:
 
 ### Windows
 
-To install the necessary compiler toolchain on Windows, run `npm install -g --production windows-build-tools`. If you're using a Node version older than 8.0, you must use `windows-build-tools` version 3.0 (not the newest version, 4.0).
+To install the necessary compiler toolchain on Windows, run `npm install -g --production windows-build-tools`. If you're using a Node version older than 8.0, you must use `windows-build-tools` version 3.0 (not 4.0, the newest version).
 
 ### macOS
 
-MacOS ships with the `clang` compiler suite. Use `clang --version` to ensure that you have the compiler installed. If clang is **not** already installed, run `$ xcode-select --install` to install it.
+macOS ships with the `clang` compiler suite. Use `clang --version` to ensure that you have the compiler installed. If Clang is **not** already installed, run `$ xcode-select --install` to install it.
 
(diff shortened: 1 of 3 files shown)


content/installation/node/NodeInstall.md
```diff
@@ -4,6 +4,31 @@ description: "Installing the Node.js Agent"
...
+> **Notes:** 
+ * If you can build other modules with C++ bindings like [node-sass](https://www.npmjs.com/package/node-sass) and [sqlite3](https://github.com/mapbox/node-sqlite3), you probably meet all of the requirements for Contrast.
+ * The official [Node Docker images](https://hub.docker.com/_/node/) come with all of the prerequisites installed.
+
+### Windows
+
+To install the necessary compiler toolchain on Windows, run `npm install -g --production windows-build-tools`. If you're using a Node version older than 8.0, you must use `windows-build-tools` version 3.0 (not the newest version, 4.0).
+
+### macOS
+
```
kaitlynmortimer: @Avuxo It looked like all of these (Windows, macOS, etc.) are supposed be subsections under Prerequisites. If that's wrong, I can readjust the subhead levels.

content/installation/node/NodeInstall.md
```diff
@@ -50,3 +75,7 @@ First, add the following script to your application's *package.json*:
 ```
 
 Then, the agent can be run with ```npm run contrast```. This npm script can be changed to include other runtime configurations, such as an alternate configuration file location. For more information, see [Node Agent Configuration](installation-node.html#node-config).
+
+## Errors during Installation
+
+`Invalid ELF Header` - This error occurs when a native C++ module targeting Linux has been compiled for a non-Linux target. Ensure that you are installing the agent from a compatible system.
```
kaitlynmortimer: @Avuxo Should we add this bit to https://docs.contrastsecurity.com/troubleshooting-node.html#node-ts-install instead?

content/installation/node/NodeInstall.md
Avuxo: Yep, that was an oversight on my part.

content/installation/node/NodeInstall.md
Avuxo: I'll move it over there; good catch.

meandthemoon:
Would it make sense to document here that the auto-update feature will ignore updates for major version bumps?

PR #597, opened 2019-01-15
Update NodeSupportedTechnologies.md

j0nS3idman:
Add link to working with transpilers from supported technologies page (for Node).


kaitlynmortimer:
Thanks for the suggestion. I reformatted this change in https://github.com/Contrast-Security-OSS/docs/pull/603.

PR #606, opened 2019-02-14
Update library analysis description

rstatsinger:
Add description of how we calculate the library letter grade


kaitlynmortimer:
@rstatsinger - Thanks for the suggestion! We have this info in https://docs.contrastsecurity.com/user-libraries.html#analysis. I created PR https://github.com/Contrast-Security-OSS/docs/pull/607 to make this more clear.

PR #612, opened 2019-03-01
🔥 External rules have been removed from the product for a while

gilday:
We removed these External Rules from the product a while ago, so I don't think we should host these docs. These _could_ be useful to a user that has a _really_ old version of our software and is using these rules, but as far as we know nobody still uses this.


kaitlynmortimer:
Can we move this file to an Archive folder instead of deleting it?

kaitlynmortimer:
Also, I'll need to merge https://bitbucket.org/contrastsecurity/open-docs/pull-requests/253/remove-article-rename-remaining-one/diff (with the necessary changes to the page structure) immediately after you guys merge this.

gilday:
> Also, I'll need to merge https://bitbucket.org/contrastsecurity/open-docs/pull-requests/253/remove-article-rename-remaining-one/diff (with the necessary changes to the page structure) immediately after you guys merge this.

Oh man I hate these git submodules... makes everything so needlessly complex

gilday:
> Can we move this file to an Archive folder instead of deleting it?

Sure. I am not familiar with how the Archive works. Do archived articles still show up in search results? Can users navigate to the archive? How does one put something in the archive?

kaitlynmortimer:
It's just a folder named "archive" (i.e., content/admin/eop_maintenance/upgrade/archive) to which we move outdated/unwanted content for that subtopic. It shouldn't show up in search results. If the users could navigate to it...I'm not sure how; but, Mike could probably tell you for sure.

gilday:
@miketmoore does this archive directory get deployed to the site? Another way of asking this is, "can users see any content in the archive, or is it just there for our records"?

miketmoore:
@gilday Not by default - the content (markdown files) inside this repo must be specified explicitly in the site generator to be visible in the published site. I don't see this anywhere, but the archive directories (at least one) appear to be linked directly in some places on the published site - I see this on https://docs.contrastsecurity.com/release.html#360 which links to https://github.com/Contrast-Security-OSS/docs/tree/master/content/release/archives

gilday:
OK, so if something is in the archive, then it won't show up in the navigation, but you can still link to that content from old articles like release notes so that we don't dead those links - is that the idea?

kaitlynmortimer:
> OK, so if something is in the archive, then it won't show up in the navigation, but you can still link to that content from old articles like release notes so that we don't dead those links - is that the idea?

@gilday The hyperlinks would still be dead. But, we always keep this stuff for reference in case someone has a specific question about it, we want to redo the article one day, etc.

gilday:
I see. I feel like git already gets you that because it's going to have a record of everything you delete, but if that's the process that works for us that's fine. I'll redo this PR with a move instead of a delete

PR #618, opened 2019-03-19
Remove 1 App - 1 JVM release note

scorlett:
As the move to 1 App - 1 JVM has been pushed back for the time being, this release note should be removed to avoid any confusion.

### DIFF
--- a/content/release/ReleaseNotes359.md
+++ b/content/release/ReleaseNotes359.md
@@ -33,13 +33,7 @@ Our gift to you: completely revamped grids for all your applications, servers an
 
 ### Java summary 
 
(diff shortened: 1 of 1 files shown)


kaitlynmortimer:
@scorlett @DanFiedler 

Are the properties no longer being deprecated? Is that part no longer relevant or helpful?

DanFiedler:
> @scorlett @DanFiedler
> 
> Are the properties no longer being deprecated? Is that part no longer relevant or helpful?

We're still figuring out the exact deprecation plan for the older properties. It is safer to remove the mention of deprecating properties to avoid confusion.

PR #620, opened 2019-03-20
Adding fixed-width templates

meandthemoon:
The maximum line length will be 80 characters.
Within each (sub)section, there is now a path in the header that shows where users are within the spec document, e.g. `api.certificate`, `assess.sampling`.


kaitlynmortimer:
Looks good!

PR #627, opened 2019-03-29
:memo: CONTRAST-31541 - replace contents of auto generated documentat…

rszewczyk:
…ion with a message about how to suggest changes to it


rszewczyk:
@kaitlynmortimer after you approve this we'll need to wait until the first version of open docs with the auto generated templates is deployed (so that we don't actually deploy these placeholder docs)

PR #634, opened 2019-04-03
Update .NET agent config with common config options

DanFiedler:
- Update .NET agent config documentation to be based on common config
- Add details on how to configure via environment variables or via web.config
- Add note about Legacy config options, users should prefer yaml-based config
- Update and add a few missing config options
- Add examples of commonly desired yaml-based configuration

### DIFF
--- a/content/installation/net/configuration/YAML-properties-net.md
+++ b/content/installation/net/configuration/YAML-properties-net.md
@@ -15,8 +15,8 @@ Configuration values use the following order of precedence:
 1. Corporate rule (e.g., expired license overrides `assess.enable`)
 2. Specific environmental variable
 3. Generic environment variable value
-4. Application-specific configuration file value (i.e. web.config)
-5. User configuration file value (i.e. contrast_security.yaml)
+4. Application-specific configuration file value (i.e. *web.config*)
+5. User configuration file value (i.e. *contrast_security.yaml*)
 6. Contrast UI value
 7. Default value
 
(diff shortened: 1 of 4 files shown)


content/installation/net/configuration/YAML-properties-net.md
```diff
@@ -26,6 +27,28 @@ The *contrast_security.yaml* file should be placed on the file system using one
 * Specify the path to the YAML file with the environment variable `CONTRAST_CONFIG_PATH`.
 * Place the *contrast_security.yaml* file in the data directory specified during agent install. (The default location is * %ProgramData%\Contrast\dotnet\*. As a result, the default file path would be *%ProgramData%\Contrast\dotnet\contrast_security.yaml*.)
 
+## Using Environment Variables
+Every configuration option supported by the *contrast_security.yaml* file can also be specified using environment variables. 
+
+Environment variable names are derived from the yaml path by replacing path segment delimiters (`.`) with double underscores (`__`) and prefixing the result with `CONTRAST__` e.g. `server.name` becomes `CONTRAST__SERVER__NAME` while `api.api_key` becomes `CONTRAST__API__API_KEY`.
```
Silvenga: We need to make sure that these underscores are rendered correctly. Github markdown renders correctly, but the markdown engine we are using for docs doesn't handle them well.

I think this is fine as we are using the ticks.

ldemidov: Prefixing with CONTRAST__ is not required.  Should we change this text?

content/installation/net/configuration/YAML-properties-net.md
```diff
@@ -226,4 +254,47 @@ Use the properties in this section to set metadata for the server hosting this a
...
+    enable_chaining: true
+```
+
+### Whitelist an Application Pool for Instrumentation
+The following configuration excludes all application pools except `ExampleAppPool` from instrumentation by the .NET Agent. This can help improve performance on applications you do not wish to analyze.
+
+```
+agent:
+  dotnet:
+    app_pool_whitelist: ExampleAppPool
```
Silvenga: Maybe give an example with a list?

content/installation/net/configuration/YAML-properties-net.md
Silvenga: Are we sure? `EnvironmentConfigProvider` seems to be unconditionally prefixing with `CONTRAST__`.

ldemidov: yea this code is admittedly hard to read but that's the case

kaitlynmortimer:
@DanFiedler Based on the changes in the YAML Properties article, can we update the YAML template as well?

DanFiedler:
@kaitlynmortimer Yes, I'll do that now.

PR #638, opened 2019-04-08
Contrast 25672 glossary

kaitlynmortimer:


content/glossary/Glossary.md
```diff
@@ -1,145 +1,345 @@
...
 
-Each **event** contains the runtime values of the underlying *object*, the parameters to the method invoked, the return from the method, and the runtime stack trace.
+Each **event** contains the runtime values of the underlying object, the parameters to the method invoked, the return from the method and the runtime stack trace.
 
 ---
 
 ## Duplicate Trace Elimination
 
-The Contrast *engine* will detect the same vulnerability conditions over and over in an application, and this would normally cause the same *trace* to be generated repetitively. To address this, there is **duplicate trace elimination** logic in both the *engine* and the *TeamServer* to eliminate duplicate *traces*.
+The Contrast **engine** detects the same vulnerability conditions over and over in an application; this would normally cause the same **trace** to be generated repetitively. To address this, there is **duplicate trace elimination** logic in both the **engine** and the **Contrast UI** to eliminate duplicate traces.
```
kaitlynmortimer: @DanFiedler We use the term "engine" in some old definitions, but the term itself is never defined. Is this synonymous with "agent"?

content/glossary/Glossary.md
```diff
@@ -1,145 +1,345 @@
...
-The **Contrast site** is a web application that receives *traces* from Contrast *engines* and allows the analyst to view and triage the results.
+The **Contrast UI** is the web interface from which users download agents and view information on their **applications** and reported **vulnerabilities**. This also includes the REST API to which the **agents** report data.
 
 ---
 
 ## Creation Event
 
-When an *untrusted data source* provides data to the application, a **Creation Event** (or **Source Event**) is generated that documents what occurred. This information is stored with the *tracked* data object that is created and is used to describe where untrusted data came from if a *rule* generates a *trace*.
+When an **untrusted data source** provides data to an application, Contrast generates a creation event (or source event) that documents what occurred. Contrast stores this information with the **tracked** data object that it created, and uses it to describe the source of the untrusted data, if a **rule** generates a **trace**.
+
```
kaitlynmortimer: @DanFiedler We don't use the term "term" anywhere else in the docs (or the UI, as far as I know). Can we update terms to "vulnerability"?

content/glossary/Glossary.md
hmcginnis: **trace** was the old term for a vulnerability / vulnerable data path, not sure if we're still using it or not.
**engine** was the old term for agent. I think it should be replaced.

PR #641, opened 2019-04-11
✨ CONTRAST-32204 Add the Gradle, Docker, Maven Central Example

gilday:
Shows a Gradle user running integration tests with Docker how to integrate the Contrast Java agent from Maven Central.


gilday:
@kaitlynmortimer I updated the version number in this article and it is now ready to ship

PR #652, opened 2019-04-30
Remove or rewrite info re: request time

rstatsinger:
Remove or rewrite this line: :"In typical applications, Contrast adds around 2x round-trip time to a request that contains a lot of business logic with all the bells and whistles turned on."  It's 4 years old and does not seem to be current. Synch with engineering and replace with something more current and accurate, or at least, something more palatable to customers.


kaitlynmortimer:
@phil-mcgee @rszewczyk 

I'm adding both of you for input as the on-call for Java. 

Can this line simply be removed, or should I open a ticket for more extensive updates?

PR #655, opened 2019-05-02
Add net core profiler chaining steps

m0nkey653:
Added profiler chaining instructions under the Net Core Installation instructions.


DanFiedler:
Looks good. I wonder if a final example of a complete set of environment variables would also be helpful.

m0nkey653:
> Looks good. I wonder if a final example of a complete set of environment variables would also be helpful.

Added the complete set of environments for both New Relic and AppD

PR #658, opened 2019-05-03
Change key of node agent update property; Correct table alignment

gibbsjosh:
Line 28 is the only content change here. `--agent.auto_update` is now `--agent.auto_update.enable` to reflect deprecation of the earlier property name. When making this change, I noticed the table had become misaligned in a previous commit, so I corrected that, too.


cixel:
@bizob2828 in case you were going to take care of this

bizob2828:
I addressed this as well as many others in #659. I'm going to close this one

PR #676, opened 2019-06-03
CONTRAST-34186 :pencil: updating protect exclusion documents to inclu…

argonotix:
Updated Protect exclusions document to include tables of:

- Sample Input exclusions
- Sample URL exclusions
- "Protect Rule Exclusion Support Matrix"
- Regular Expression quick reference

I'll leave it up to reviewer comments, but I feel that any rules that are governed only by URL exclusion should be removed from the 'support matrix' table.


content/admin/policy_mgmt/Exclusions.md
```diff
@@ -84,4 +107,62 @@ Alternatively, you can see a global list of existing exclusions across all appli
 
 <a href="assets/images/Application-exclusions.png" rel="lightbox" title="Manage exclusions in Policy Management"><img class="thumbnail" src="assets/images/Application-exclusions.png"/></a>
 
-
+## Protect Rule Exclusion Support Matrix
+
+All rules support URL exclusions in Protect mode.  If a rule is not listed here, it can be assumed that the rule may be excluded using URL exclusions.
```
gilday: Do we need other teams to add their language specific Protect rules to this list?

argonotix: That'd be my assumption as well.

seschis: why do only parts of input exclusions work on certain rules? Ex. Bean Introspection RCE Vulnerability can use Parameter but not Header? Won't this list be hard to maintain as new rules are added or existing rules change? Is there any way to generalize the description?

kaitlynmortimer:
@seschis Do you have any more feedback on the updated PR?

PR #677, opened 2019-06-10
Update YAML-properties-net.md per ZenDesk-12713

pprofili-contrast:
The two example on using auto update incorrectly show a hyphen rather than an underscore.

### DIFF
--- a/content/installation/net/configuration/YAML-properties-net.md
+++ b/content/installation/net/configuration/YAML-properties-net.md
@@ -284,7 +284,7 @@ The following configuration disables the auto-update feature that automatically 
 
 ```
 agent:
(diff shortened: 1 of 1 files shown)


kaitlynmortimer:
@DanFiedler Can you please provide the most-recent version of the .NET template, if this one is out of date?

pprofili-contrast:
I believe the template is good (at least it is for auto_update) but I will let Dan confirm.

DanFiedler:
This looks good. The only issue I see is that we need to update the valid values for tls_version:
- tls
- tls11
- tls12
- (tls|tls11)
- (tls11|tls12)
- (tls|tls11|tls12)

PR #689, opened 2019-06-22
Reference the supported LTS status for node versions 

palmergs:
This PR drops expicit node version number in the supported technologies page for Node and references the specific LTS support status that we support.


SteveFeldman:
Approved

PR #691, opened 2019-06-25
Clarify Contrast's effect on application performance

rstatsinger:
That paragraph is a destroyer of deals and gives our competitors ammunition. It should be rewritten

### DIFF
--- a/content/troubleshooting/java/application/Java-EffectsOnAppPerformance.md
+++ b/content/troubleshooting/java/application/Java-EffectsOnAppPerformance.md
@@ -13,7 +13,6 @@ When you start your server with Contrast, you'll see a few messages that indicat
 
 ## Request Processing Time
 
-It's probably more important to think about how Contrast affects the round-trip time. In typical applications, Contrast adds around 2x round-trip time to a request that contains a lot of business logic with all the bells and whistles turned on. Round trip times for static resources typically don't get measurably worse. In requests where the total round-trip time is dominated by database or Web Service calls, Contrast's effect will be less noticeable.
 
 If better performance is really important to your environment, consider the following options:
 


content/troubleshooting/java/application/Java-EffectsOnAppPerformance.md
```diff
@@ -13,7 +13,6 @@ When you start your server with Contrast, you'll see a few messages that indicat
 
 ## Request Processing Time
 
-It's probably more important to think about how Contrast affects the round-trip time. In typical applications, Contrast adds around 2x round-trip time to a request that contains a lot of business logic with all the bells and whistles turned on. Round trip times for static resources typically don't get measurably worse. In requests where the total round-trip time is dominated by database or Web Service calls, Contrast's effect will be less noticeable.
```
gilday: There's some helpful information in here we should consider keeping to help users know what to expect.

How do you feel about changing this paragraph to

> Contrast's affect on application performance varies depending on the data processing characteristics of the application. Applications whose request processing time is dominated by IO operations (e.g. database or web service calls) are less likely to perceive Contrast's effects. On the other hand, Contrast's sensors will be more active in an application that performs unusually large text processing in memory.

content/troubleshooting/java/application/Java-EffectsOnAppPerformance.md
```diff
@@ -13,7 +13,6 @@ When you start your server with Contrast, you'll see a few messages that indicat
...
 
 If better performance is really important to your environment, consider the following options:
```
gilday: > If better performance is really important to your environment

This seems like a silly thing to say, because performance is always important. Instead, what if we reword this to say

> Contrast exposes a number of performance tuning configurations to help users make trade-offs appropriate for their environments. Consider the following tuning options:

content/troubleshooting/java/application/Java-EffectsOnAppPerformance.md
rstatsinger: Remove this:

If better performance is really important to your environment, consider the following options:

Replace with:

Here are a set of performance tuning suggestions for you.

Please keep in mind: our primary goal is to sell software. Our secondary goal is to explain how our product works, but only in the service of our primary goal.

rstatsinger: As you might expect, Contrast's analysis does make your app run a little slower. The good news is it's generally not enough for anyone to complain about, and the results are definitely worth it.
 
THIS IS HORRIBLE! REMOVE IT. WE ARE TRYING TO SELL SOFTWARE.

PR #703, opened 2019-07-03
Python-install: update angle brackets to square brackets

ghost:
Last week I submitted a PR that updated the Python installation instructions. I noticed today that the markdown wasn't being rendered properly due to my use of angle brackets.

This PR changes the angle brackets to square brackets.

Is there an easy way for me to launch the site locally so I can review my changes in rendered form before they become live? I'm still not totally sure that this change will fix the issue.

### DIFF
--- a/content/installation/python/PythonInstall.md
+++ b/content/installation/python/PythonInstall.md
@@ -7,20 +7,20 @@ tags: "python agent installation"
 
 To install the Contrast agent into your Python application, you must complete the following steps.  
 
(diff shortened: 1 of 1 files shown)


kaitlynmortimer:
> Is there an easy way for me to launch the site locally so I can review my changes in rendered form before they become live? I'm still not totally sure that this change will fix the issue.

@miketmoore Can you advise?

kaitlynmortimer (approved):

PR #706, opened 2019-07-08
CONTRAST-35869 Ruby supported tech

hmcginnis:
:pencil: update Ruby supported techs.

We've moved our documentation to match the style laid out by the Node team and to make our LTS support both more explicit and dynamic. In addition, we've moved the MuslC work around to its own page. 

We'd like `content/installation/ruby/RubyMuslC.md` to be mapped to `installation-ruby.html#muslc`

### DIFF
--- a/content/installation/ruby/RubyMuslC.md
+++ b/content/installation/ruby/RubyMuslC.md
@@ -1,0 +1,30 @@
+<!--
+title: "Working with MuslC"
+description: "Notes on Contrast compatibility with the MuslC compiler"
+tags: "ruby agent compatibility muslc"
+-->
+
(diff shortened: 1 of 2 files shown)


kaitlynmortimer:
@hmcginnis Do you want the MuslC article on the same page with the existing Supported Technologies article?

hmcginnis:
> @hmcginnis Do you want the MuslC article on the same page with the existing Supported Technologies article? The structure would be like the Ruby > Configuration page.

@kaitlynmortimer, I was hoping it could be it's own page, like how Node did it for their Working with Transpilers link in https://docs.contrastsecurity.com/installation-node.html#node-supported

PR #709, opened 2019-07-11
Java server copy edits

kaitlynmortimer:


evamill:
@gilday These docs changes look mostly like style updates to the Server Configuration section of the Open Docs for Java. They have older dates on them, but Kaitlyn set them to merge before leaving last week. They look fine to go, in my opinion. Who can make sure these go out?

gilday:
@evamill anyone familiar with the Open Docs project should be able to deploy these changes. We will need a corresponding change in the open-docs Bitbucket repository. I see Kaitlyn has such a change there from July, but it is 164 commits behind. Because the changes are so old, I am hesitant to blindly deploy these. I will work with the Java team to get these deployed next sprint.

evamill:
I've been asking to have access to docs in BitBucket for a week--can't see them. No one seems to be able to help me with that, despite many requests to IT and some to engineering. In the notes Kaitlyn left last week, she didn't explain why/when we would use BitBucket vs GitHub or what the relationship is between the two. So, I'm in the dark on this for two reasons, then.

terra4383:
Updated this branch with the latest from master and verified it could be merged with @gilday

PR #710, opened 2019-07-12
Remove serverId maven parameter

andersonshatch:
The maven plugin does not have a `serverId` parameter, per [its readme](https://github.com/Contrast-Security-OSS/contrast-maven-plugin/blob/master/README.md#configuration-options).


gmoore:
Thanks @andersonshatch

PR #715, opened 2019-07-22
✨ CONTRAST-36484 contrast-java-agent-exec-helper docs

gilday:
Adds page "Host Installation" with two articles

1. Red Hat Hosts
1. Ubuntu Hosts

Adds a troubleshooting page for the contrast-java-agent-exec-helper.

I've deployed the contrast-java-agent-exec-helper to production, so these changes are all that's left to release this feature.


kaitlynmortimer:
@scorlett Just a heads up about the new Troubleshooting article. Should we send you the markdown file directly?

scorlett:
Yes please. The Zendesk KB doesn't use markdown but that will at least show me how it's intended to be formatted and reproduce.

PR #716, opened 2019-07-23
Dotnet doc updates

Silvenga:
A couple of updates:
- New installer options
- Documented debug options.

### DIFF
--- a/content/installation/net/install/Installer-CommandLine-Options.md
+++ b/content/installation/net/install/Installer-CommandLine-Options.md
@@ -1,0 +1,38 @@
+<!--
+title: "Contrast .NET Agent Installer Command Line Options"
+description: "Contrast .NET agent installer command line options"
+tags: "installation agent .NET windows silent unattended installer setup"
+-->
+
+## .NET Agent Installer Command Line Options
+
+The .NET agent installer supports additional options when invoked from the command line.
+
+### Silent Installation
+
+The .NET agent installer supports standard silent installation options. These options tell the installer to run silently and unattended, which means that the installer don't require your interaction and don't present the installer's UI.
+
+- Install: `ContrastSetup.exe -s -norestart`
+
+### Uninstalling and Repairing
+
+The .NET agent installation can be uninstall or repaired using standard Windows features (e.g. the Programs and Features Control Panel, Powershell, etc.). However, it may be desirable to use the .NET agent installer to perform these actions instead.
+
+**Attended:**
+- Uninstall: `ContrastSetup.exe -uninstall`
+- Repair: `ContrastSetup.exe -repair`
+
+**Unattended:**
+- Uninstall: `ContrastSetup.exe -uninstall -s -norestart`
+- Repair: `ContrastSetup.exe -repair -s -norestart`
+
+### Custom Installation
+
+The .NET agent installer supports several additional options accessible when installing using the command line. Supported options are shown below:
+
+| Option | Description | Example |
+|--|--|--|
+| `StartTray` | When set to `0`, suppress starting the Tray application when installation is completed. This is recommended when installing on Windows Server Core instances. Defaults to `1`. | `StartTray=0` |
+| `PathToYaml` | Specify a custom YAML configuration file. Defaults to a file called `contrast_security.yaml` located relative to the installer's location. | `PathToYaml=c:\contrast_security.yaml` |
+| `SERVICE_STARTUP_TYPE_MANUAL` | When set to `1`, and when installing/upgrading/repairing, set the Contrast service startup type to Manual. This _must_ be provided on installs/upgrades/repairs. Defaults `0` (Automatic Delayed Start). | `SERVICE_STARTUP_TYPE_MANUAL=1` |
+| `SUPPRESS_SERVICE_START` | When set to `1`, and when installing/upgrading/repairing, suppress automatically starting the service. This _must_ be provided on installs/upgrades/repairs. Defaults `0`. | `SUPPRESS_SERVICE_START=1` |
--- a/content/installation/net/install/Installer-CommandLine-Options.md
+++ b/content/installation/net/install/Installer-CommandLine-Options.md
@@ -1,3 +1,4 @@
+
 <!--
 title: "Contrast .NET Agent Installer Command Line Options"
 description: "Contrast .NET agent installer command line options"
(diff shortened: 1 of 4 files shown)


content/installation/net/install/Installer-CommandLine-Options.md
```diff
@@ -0,0 +1,38 @@
+<!--
```
DanFiedler: I think adding this new file requires a structural change by @kaitlynmortimer . I think it is worth breaking the installer options content out into its own file from the general installer documentation but I know it requires a coordinated change to a separate repo...

Silvenga: Oh, was wonder how that worked.

content/installation/net/configuration/YAML-properties-net.md
```diff
@@ -288,6 +289,20 @@ agent:
     enable: false
 ```
 
+### Troubleshooting Options
```
DanFiedler: We don't *have* to document these settings here. There are pros and cons to documenting these settings here. 

We do need to update this article: https://github.com/Contrast-Security-OSS/docs/blob/master/content/troubleshooting/net/GettingNetAgentLogs.md

Silvenga: Ooh, I understand what you were wanting now.

Silvenga: Fixed.

content/installation/net/install/Installer-CommandLine-Options.md
```diff
@@ -0,0 +1,38 @@
...
+- Uninstall: `ContrastSetup.exe -uninstall`
+- Repair: `ContrastSetup.exe -repair`
+
+**Unattended:**
+- Uninstall: `ContrastSetup.exe -uninstall -s -norestart`
+- Repair: `ContrastSetup.exe -repair -s -norestart`
+
+### Custom Installation
+
+The .NET agent installer supports several additional options accessible when installing using the command line. Supported options are shown below:
```
ldemidov: I think an example of setting one of these would be nice, to show the syntax.

content/installation/net/install/Installer-CommandLine-Options.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Silent Installation
+
+The .NET agent installer supports standard silent installation options. These options tell the installer to run silently and unattended, which means that the installer don't require your interaction and don't present the installer's UI.
+
+- Install: `ContrastSetup.exe -s -norestart`
+
+### Uninstalling and Repairing
+
+The .NET agent installation can be uninstall or repaired using standard Windows features (e.g. the Programs and Features Control Panel, Powershell, etc.). However, it may be desirable to use the .NET agent installer to perform these actions instead.
```
ldemidov: "may be desirable" sounds a bit vague.  I would just say something like "However for automation and scripted scenarios you may also perform these actions from the command line".

content/installation/net/install/Installer-CommandLine-Options.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Attended:**
+- Uninstall: `ContrastSetup.exe -uninstall`
```
ldemidov: Lets add one for "Install" too.
Do we need repair?  I don't know if that's widely used or people know what it does.

content/installation/net/install/NetInstallation.md
```diff
@@ -29,33 +29,23 @@ You can fully configure the agent using the *contrast_security.yaml* file. See t
...
-- Creates the specified data directory for the agent that's primarily used to store agent log files and configuration (e.g., *C:\ProgramData\Contrast\dotnet*). 
+- Places the agent’s files on a disk in the specified install location (e.g., *C:\Program Files\Contrast\dotnet*). This includes several dynamic link libraries (DLLs) and executables, such as the background Windows service that drives agent behavior.
+- Creates the specified data directory for the agent that's primarily used to store agent log files and configuration (e.g., *C:\ProgramData\Contrast\dotnet*).
 - Registers the agent’s background Window service with the operating system.
-- Starts the agent’s background Windows service and Tray (UI) application. This service has a number of responsibilities: 
-  - Preparing the environment for instrumentation by registering the agent’s profiler component with IIS through environment variables, and restarting IIS. This causes the CLR to load the agent’s profiler, which is responsible for instrumenting analyzed applications. 
+- Starts the agent’s background Windows service and Tray (UI) application. This service has a number of responsibilities:
```
ldemidov: Should we point out the background service name here?  It might be worth mentioning that the agent can be restarted by restarting the service: this is something we've recommended to people for a while as a foolproof way to restart the agent.  Not sure if this the best place for it but it seems like this document is going beyound 'installation' and into 'system documentation'.

content/troubleshooting/net/GettingNetAgentLogs.md
```diff
@@ -14,22 +14,22 @@ You can change which information is logged by changing the logging level in the
 
 ## Types of Bugs
```
ldemidov: Why don't we refer to these as "Issues" instead of "Bugs"?

content/installation/net/install/Installer-CommandLine-Options.md
Silvenga: Install, you mean the Silent Installation section :D 

I like keeping repair since it's stating we do support standard MSI flags. We have to support it anyway because we can't prevent users from using that feature, so we might as well advertise support.

content/installation/net/install/NetInstallation.md
Silvenga: I really want to update the agent service name before documenting it. 😹 

We could recommend using the those shortcuts to restart the service instead? I'm not sure if those are documented anywhere.

content/installation/net/install/Installer-CommandLine-Options.md
ldemidov: I think it would read better with all 3 options in the same table and using the same terminology.  The "silent" installation is the same thing as unattended here, and there is no command for "attended".

Silvenga: The lack of a silent flag is attended. In a lot of installers, most of Microsoft's installers, unattended != silent. I just want to make it plain that you can use some of the flags with using the UI to install e.g. the service startup flag can be used with the UI.

kaitlynmortimer: @Silvenga Is this resolved?

Silvenga: Yes, I merged the install/uninstall sections.

PR #731, opened 2019-08-19
CONTRAST-37441 Agent YAML configuration

hmcginnis:
:truck: rename generated files for consistency
:pencil: update files to say they're autogenerated and replaced


hmcginnis:
I can't / won't merge this one until the update to the generation script is merged so that the .NET Core configuration file is properly updated.

kaitlynmortimer:
@hmcginnis Don't we still want the templates available for people to copy from OD?

hmcginnis:
Deployed the generation script. Updates to YAML are now live. 

These changes are now just to make sure we don't get back in this state. The contents of these files are not reflected on the live Open Docs site as the build script replaces them.

PR #734, opened 2019-08-22
Add official yarn support

bizob2828:
In preparation for node agent 2.6.0 I cleaned up our docs around yarn.  We no longer rely on python or compilers to install the agent.  However, I moved that stuff to the troubleshooting area.  I also added KoaTestBench as it was missing from sample apps and renamed NodeTestBench to ExpressTestBench.

@kaitlynmortimer mentioned she has to do some manual work to make the new compilers.md available/linked in the troubleshooting steps

### DIFF
--- a/content/installation/node/NodeInstall.md
+++ b/content/installation/node/NodeInstall.md
@@ -4,30 +4,6 @@ description: "Installing the Node.js Agent"
 tags: "NodeJS agent installation"
 -->
 
-## Prerequisites
-
-To install the Node agent, you must have installed:
-
-* A C++ compiler toolchain (e.g., Clang, GCC or MSVC) and a POSIX-compatible `make`
-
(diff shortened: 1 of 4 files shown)


content/installation/node/NodeInstall.md
```diff
@@ -38,10 +14,10 @@ npm install node-contrast-#.#.#.tgz --no-save
 ```
 This will add the agent to your *node_modules* folder without creating an entry in the dependencies list of your *package.json*.
 
-> **Note:** If you use yarn, you can't run `yarn add node-contrast-#.#.#.tgz`. Yarn has known issues with packages that contain bundled dependencies like our agent package. Run the following **after** you install yarn:
- ``` sh
- mkdir -p node_modules/node_contrast && tar zxvf node_contrast-#.#.#.tgz --strip 1 -C node_modules/node_contrast
- ```
+If you use yarn:
```
Avuxo: Could a note be added here like "with a Node agent >=2.6.0"

PR #736, opened 2019-08-23
Update config path environment variable for .NET Core

m0nkey653:
CONTRAST_CONFIG_PATH -> CONTRAST_CORECLR_CONFIG_PATH


m0nkey653:
Killing this PR, these variables need to be the same. Proposing we open up dotnet-core to a new yaml section.

PR #744, opened 2019-09-05
RUBY-166

hmcginnis:
:pencil: add recommendation for Passenger and Nginx

Add new recommendations for how to configure Passenger and Nginx with Ruby.


content/installation/ruby/server/Passenger.md
```diff
@@ -0,0 +1,66 @@
...
+title: "Installing with Passenger"
+description: "Recommendations for installing in applications run by Passenger"
+tags: "ruby agent installation Passenger"
+-->
+
+## Inability to Start the Application
+
+### Cause - Application Timeout
+
+The most common issue faced when integrating with an application started and monitored by Passenger is the increase in
```
kaitlynmortimer: @hmcginnis Is this article intended to be used primarily as troubleshooting?

hmcginnis: it's a combination of trouble shooting and installation instructions when the agent is installed in applications managed by Passenger

PR #745, opened 2019-09-06
Node PCF docs

gmoore:
https://bitbucket.org/contrastsecurity/open-docs/pull-requests/356/


content/installation/node/Node-cloud-foundry.md
```diff
@@ -0,0 +1,151 @@
...
+
+```
+cf bind-service <app_name> <name_of_service>
+```
+
+You should now see your application Contrast UI.
+
+## Contrast Service Broker Tile
+
+See [Contrast Service Broker Tile](https://docs.contrastsecurity.com/installation-javaserver.html#cloud) in the Java configuration documentation.
```
kaitlynmortimer: @gmoore Do all of the Java instructions for the tile apply? If so, I'll just copy them over.

gmoore: Yes they do but I thought it better to just keep them in one place. Otherwise, when they do change, I'll probably forget to change both.

PR #749, opened 2019-09-11
Update DebianRepository.md

secguy:
- added section where we say where the command will save the contrast.jar file. 
- added a bullet to say we required a yaml file and where it can be downloaded

### DIFF
--- a/content/installation/java/standard/DebianRepository.md
+++ b/content/installation/java/standard/DebianRepository.md
@@ -18,5 +18,8 @@ echo "deb https://pkg.contrastsecurity.com/debian-public/ all contrast" | sudo t
 ```console
 sudo apt-get update && sudo apt-get install contrast-java-agent
 ```
+The above command will install the contrast.jar file in the /opt/contrast directory in your file system. If the directory is not available, it will create it automatically. 
+
+* Once you've installed the agent file, you will need to link it to the contrast_security.yaml configuration file. You can download the yaml file from TeamServer.  
 
 Although the *jar* file you can [download from the Contrast UI](installation-javastandard.html#contrast-ui) is preconfigured with connection parameters, you must provide Contrast connection parameters using the agent [configuration properties](installation-javaconfig.html) when using the *contrast-agent.jar* from the Debian repository.


content/installation/java/standard/DebianRepository.md
```diff
@@ -18,5 +18,8 @@ echo "deb https://pkg.contrastsecurity.com/debian-public/ all contrast" | sudo t
 ```console
 sudo apt-get update && sudo apt-get install contrast-java-agent
 ```
+The above command will install the contrast.jar file in the /opt/contrast directory in your file system. If the directory is not available, it will create it automatically. 
+
+* Once you've installed the agent file, you will need to link it to the contrast_security.yaml configuration file. You can download the yaml file from TeamServer.
```
gilday: The next paragraph introduces Configuration. Let's take this out

content/installation/java/standard/DebianRepository.md
```diff
@@ -18,5 +18,8 @@ echo "deb https://pkg.contrastsecurity.com/debian-public/ all contrast" | sudo t
 ```console
 sudo apt-get update && sudo apt-get install contrast-java-agent
 ```
+The above command will install the contrast.jar file in the /opt/contrast directory in your file system. If the directory is not available, it will create it automatically.
```
gilday: specifically, `/opt/contrast/contrast-agent.jar`. Let's spell out the whole path

```
/# dpkg -L contrast-java-agent
/usr
/usr/share
/usr/share/doc
/usr/share/doc/contrast-java-agent
/usr/share/doc/contrast-java-agent/copyright
/usr/share/doc/contrast-java-agent/changelog.gz
/opt
/opt/contrast
/opt/contrast/contrast-agent.jar
```

gilday: > If the directory is not available, it will create it automatically. 

Do we need to call this out? I don't want to confuse anyone by adding unnecessary details. I think it's implied that the directories are created, because we never instructed the user to create the directory.

gilday:
oh... this is closed now?? @secguy what's the deal?

gilday:
Disregard, it's small, I'll open another https://github.com/Contrast-Security-OSS/docs/pull/750

content/installation/java/standard/DebianRepository.md
secguy: it's fine, we don't need to mention it.

PR #750, opened 2019-09-11
:sparkles: Clarify where the Java agent packages install the agent

gilday:


tylernhoward:
Looks good to me.

gilday:
I double checked in a centos and ubuntu docker container before sending this

PR #759, opened 2019-09-23
we need to tell the user about yaml file

secguy:
added a new step asking the user where to get the yaml file. 
comment about the exec helper should be just a comment, not a step to install the agent

### DIFF
--- a/content/installation/java/host/UbuntuHosts.md
+++ b/content/installation/java/host/UbuntuHosts.md
@@ -37,7 +37,9 @@ $ bash -c "java -Dcontrast.stdout=true -version 2>1 | grep Contrast | head -n 1"
 [Contrast] Wed Aug 15 17:37:23 UTC 2018 No TeamServer configuration detected. Agent will only be reporting to local listeners (e.g., Eclipse Plugin).
 ```
 
-* The Exec Helper affects all new shells, as indicated by the `bash -c` in the verification command. Consequently, it will not be present in your current shell. Reload your shell to enable the Exec Helper. Restart any `java` services to enable Contrast for those services.
+* The *contrast-agent.jar* expects a contrast_security.yaml file to be located in /etc/contrast/java/ directory. You can either [download it from the Contrast UI](installation-javastandard.html#contrast-ui) or create it and paste the required [Contrast UI properties](installation-javaconfig.html) 
+
+The Exec Helper affects all new shells, as indicated by the `bash -c` in the verification command. Consequently, it will not be present in your current shell. Reload your shell to enable the Exec Helper. Restart any `java` services to enable Contrast for those services.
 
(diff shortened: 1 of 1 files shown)


kaitlynmortimer:
@gilday If you approve the content, I'll work on incorporating the new info into the existing content.

content/installation/java/host/UbuntuHosts.md
```diff
@@ -37,7 +37,9 @@ $ bash -c "java -Dcontrast.stdout=true -version 2>1 | grep Contrast | head -n 1"
 [Contrast] Wed Aug 15 17:37:23 UTC 2018 No TeamServer configuration detected. Agent will only be reporting to local listeners (e.g., Eclipse Plugin).
 ```
 
-* The Exec Helper affects all new shells, as indicated by the `bash -c` in the verification command. Consequently, it will not be present in your current shell. Reload your shell to enable the Exec Helper. Restart any `java` services to enable Contrast for those services.
+* The *contrast-agent.jar* expects a contrast_security.yaml file to be located in /etc/contrast/java/ directory. You can either [download it from the Contrast UI](installation-javastandard.html#contrast-ui) or create it and paste the required [Contrast UI properties](installation-javaconfig.html)
```
gilday: It's not entirely accurate to say that the agent requires a YAML file, nor is it accurate to say that the agent requires the YAML file to be at a specific path, because

* users can change the path to the YAML file
* users may use Java system properties instead of a YAML file
* users may use environment variables instead of a YAML file

There's a trade-off here: if we only mention the YAML file, then the instructions are terser, but we may make life difficult for those users that would be better served by system properties and environment variables (e.g. users running their Java web application in containers).

I struggle with which direction to take, but I'm leaning towards showing users how to use the YAML file then mentioning the more in-depth [Agent Configuration](installation-javaconfig.html) page.

gilday: Instead of "Contrast UI properties", let's call that page "Agent Configuration" since that is what it is called in the left-nav

gilday: > You can either [download it from the Contrast UI](installation-javastandard.html#contrast-ui)

This link isn't accurate: the installation-javastandard.html#contrast-u section instructs users how to download the _agent_ from the UI, but it doesn't mention anything about the config file. Is it possible to link to directly to the relevant wizard page on app.contrastsecurity.com instead? If not, I would prefer to omit the link entirely so we don't confuse users

gilday: I would reword this paragraph to read

> The Contrast agent requires some configuration to connect to the Contrast UI. Users may download a prepared YAML configuration file from the Contrast UI "Add Agent" wizard. For a more in-depth guide to configuring the Contrast Java agent, see [Agent Configuration](installation-javaconfig.html).

and I would copy it to all three relevant sections (Maven, Debian, Red Hat) of the standard installation page, instead of putting it on this page.

kaitlynmortimer:
Given the amount of work recommended in the requested changes, I'm transitioning this PR to a ticket. I can then take more time to evaluate the placement and wording of additional instructions.

PR #761, opened 2019-09-24
CONTRAST-38383 - Added Pylons documentation

jleo32:


content/installation/python/PythonInstall.md
```diff
@@ -101,6 +101,16 @@ app = get_wsgi_application()
 app = ContrastMiddleware(app)
 ```
 
+#### Pylons
+
+``` python
+from pylons.wsgiapp import PylonsApp
+from contrast.agent.middlewares import WSGIMiddleware
```
peter-lazorchak: This should be `from contrast.agent.middlewares.wsgi_middleware import WSGIMiddleware`. For consistency I also think we should add `as ContrastMiddleware` (a few lines down too)

jleo32: Let me update that!

jleo32: udpated

jleo32: spelling is hard

PR #769, opened 2019-10-04
Fix env var formatting

cixel:
sorry the diff is so large, [this commit](https://github.com/Contrast-Security-OSS/docs/commit/05a1b938d18782f5569689cc817588c4e993da60) has the actual meaningful changes and the rest is just adjusting the table to the new column length


bizob2828:
I'm going to merge this because I'm about to edit this file as well

PR #770, opened 2019-10-04
DOTNET-722: dotnet tweak sampling settings

ldemidov:
Update sampling CC typo and indicate that these settings are optional and intended to override Contrast UI settings


kaitlynmortimer:
@Silvenga We need to update the template as well.

Silvenga:
@ldemidov do we still want to merge this? @evamill is managing docs for the time being.

PR #772, opened 2019-10-07
Correct mention to profiler chaining option

andersonshatch:
Reference enable_chaining rather than ProfilerChainingEnabled and fix link.
Not sure if an inline code backtick block works inside a link without being able to build.


kaitlynmortimer:
Thanks, @andersonshatch. This change is included in an open PR that will be merged soon.

PR #787, opened 2019-10-18
Update JavaAgentSystemProperties.md

davidlarsen-contrast:
Updated all the configuration options to reflect the protect setting. Not all the defend configurations can simply be rewritten with the word protect. I went into our latest agent and used ~Java -jar contrast.jar properties | grep defend_config_here to find what each defend config option matched up with the newer Protect config option.


phil-mcgee:
@davidlarsen-contrast 

Hey David, I'm sorry to inform you that this document (JavaAgentSystemProperties.md) is, in my opinion, irredeemable.  I'd like to get @rszewczyk involved in this discussion since he has in past months done a lot of work updating the behavior of the utility
`java -jar contrast.jar properties`
to reflect the preference for Common Config compatible system properties and to indicate those legacy properties which are deprecated.
Further, this document should not, in general, publicize properties defined by the agent as "hidden".  We may need to discuss exceptions for those hidden properties which have been historically published in this document and on which current customers might expect to find information here.
Rob, do you think we would be better off just replacing this document wholesale with one generated by the current agent?  (I don't believe the agent will, at present , actually generate markdown but a manual reformatting of up-to-date agent generated content doesn't strike me as an unreasonable task.)
Let's talk.

kaitlynmortimer:
Let's move this discussion to the open ticket to update the Java agent configuration docs: https://contrast.atlassian.net/browse/JAVA-566

PR #789, opened 2019-10-19
Update BehaviorOtherAgents.md

davidlarsen-contrast:
Datadog is another APM that is super popular and one we see often. If DynaTrace is going to be on this list then we should add Datadog as well. 

Also, adding in the that the start order matters in the general info section up top. We get a major performance boost if we start first and in some cases it's the only way we can make it work.


kaitlynmortimer:
I'm transitioning these changes to the CS team. Please contact @marklacasse for any follow up.

PR #790, opened 2019-10-19
Update GettingAgentLogs.md

davidlarsen-contrast:
Adding in that we can set these options in the YAML as some organizations are YAML only. Also sending in an updated screenshot with redacted info.

Please view the bellow updated image of our example logs. Redacted any sensitive info and maintained a similar size/color scheme. 

GettingAgentLogs_Updated


kaitlynmortimer:
I'm transitioning these changes to the CS team. Please contact @marklacasse for any follow up.

PR #793, opened 2019-10-28
removed app performance article

kaitlynmortimer:


marklacasse:
> To where should I direct users looking for help with performance?

https://support.contrastsecurity.com/hc/en-us/articles/360000500286

PR #796, opened 2019-10-29
Contrast 38521 docs task extension

bryanateoan:
@kaitlynmortimer 
Summary of changes

- Removed Configuration for Release Gate
- Added Configuration for Task
- Added Configuration for Task as a Yaml Build Pipeline


content/tools/extensions/AzureDevOpsExtension.md
```diff
@@ -1,71 +1,152 @@
...
+## Setup and Configuration for a Task
+
+Complete the following steps to configure a task in your Azure DevOps extension. This task can be used in only an **agentless job**.
+
+### Step one: Enter Edit mode
+
+The task can be used in a release or build pipeline. Complete the appropriate steps for the pipeline you're using. 
+
+#### Release Pipeline
+
```
kaitlynmortimer: How do they enter edit mode? Will that be obvious to the reader?

content/tools/extensions/AzureDevOpsExtension.md
```diff
@@ -1,71 +1,152 @@
...
+* Enter **Edit** mode for the release pipeline you wish to add the task.
+* Select a stage for which you wish to add the task.
 
-## Configuration for Release Gate
+<a href="assets/images/AzureDevOps_release_choose_stage.png" rel="lightbox" title="Azure DevOps Release Choose Stage"><img class="thumbnail" src="assets/images/AzureDevOps_release_choose_stage.png"/></a>
```
kaitlynmortimer: We usually don't edit the images to add red indicators. Do you have the original copies of the images that we can use instead?

content/tools/extensions/AzureDevOpsExtension.md
```diff
@@ -1,71 +1,152 @@
...
-* Proceed to your severity counters, where you must set the maximum number of vulnerabilities allowed per severity. If your selected application has more vulnerabilities than allowed for that severity level, your gate will fail.
+* Proceed to your severity counters, where you must set the maximum number of vulnerabilities allowed per severity. If your selected application has more vulnerabilities than allowed for that severity level, your task will fail.
 
 <a href="assets/images/AzureDevOps_gate_part3.png" rel="lightbox" title="Azure DevOps Gate Part 3"><img class="thumbnail" src="assets/images/AzureDevOps_gate_part3.png"/></a>
 
-> **Note:** Remember that your gates evaluation settings affect all the gates on the current stage for the pre- or post-deployment conditions.
+### Step five: Set job dependency
+
+> For **Build Pipelines** - If you want to prevent the execution of a job if the task fails, you must set the job to depend on the agentless job that includes the Contrast task.
+
```
kaitlynmortimer: Is step 5 only for build pipelines? Or is this a note in case that's what the reader wants to do?

content/tools/extensions/AzureDevOpsExtension.md
bryanateoan: There is a big Edit button when the user is selecting a build pipeline. I can add an image of it if you would like?

content/tools/extensions/AzureDevOpsExtension.md
bryanateoan: I edited them right after taking the screen shots so I do not have the originals. I will retake new ones. I reused the same images for the README and Marketplace Overview, would you like me to change those as well with images without red indicators?

content/tools/extensions/AzureDevOpsExtension.md
bryanateoan: Yes, this is for build pipelines only. The release pipeline doesn't have this option. How should I indicate that it is for Build Pipelines only? Something like 
```
### Step five: Set job dependency (Build Pipelines only)
```
?

content/tools/extensions/AzureDevOpsExtension.md
bryanateoan: The images for the Release Gate has the red indicators to clarify where the user should be clicking which is what made me add the indicators for the task. Is that no longer the practice?

bryanateoan: Would you like me to replace the images with images without red indicators?

kaitlynmortimer: @bryanateoan Yes, please.

content/tools/extensions/AzureDevOpsExtension.md
bryanateoan: @kaitlynmortimer fix commited

PR #821, opened 2019-12-04
:pencil: update Ruby config docs

hmcginnis:
Update the documentation for Ruby based on questions from a recent customer installation

### DIFF
--- a/content/installation/ruby/RubyConfig.md
+++ b/content/installation/ruby/RubyConfig.md
@@ -10,24 +10,42 @@ The Ruby agent and service use a YAML file to alter the agent behavior. Go to th
 
 The configuration file is called *contrast_security.yaml* wherever it's located. The Ruby agent and service load the configuration YAML from the following paths in order of precedence (where 1 is the highest):
 
-1. The current working directory (e.g., *./contrast_security.yaml*)
-2. A subdirectory called *config*, which is the default for Ruby on Rails applications (e.g., *./config/contrast_security.yaml*)
-3. Within the server's *etc/contrast* directory (e.g. */etc/contrast/contrast_security.yaml*)
-3. Within the server's *etc* directory (e.g., */etc/contrast_security.yaml*)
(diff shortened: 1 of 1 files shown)


evamill:
@hmcginnis Just a quick overview: are these changes mostly about style and clarity? Or do they correct wrong information? Asking because it affects priorities for us. We are focused on new articles that need to go out for the December release. Copying @terra4383

hmcginnis:
@evamill , this answers questions raised by customers and brings us in line with other agent documentation. Our installation process isn't the most idiomatic (it's been a pain point for multiple customers), so this helps to alleviate that.

evamill:
I ask because the changes look lengthy and would need more review. If they can wait for the next docs release, that'd be better for our current situation as we on-board new docs people. I know our agent onboarding is problematic, but is the current Ruby documentation inaccurate (has wrong information)? Or just not as good as we would like?

hmcginnis:
I'm not sure I follow. I was under the impression that Docs were released each morning if an update is available, that they're not tied to any other product's release.

For what it's worth, almost all of this content was pulled from https://github.com/Contrast-Security-OSS/docs/blob/master/content/installation/net/configuration/YAML-properties-net.md. There shouldn't be much new content for this page.

terra4383:
@hmcginnis I'm seeing the "Logger" section under "Contrast agent properties" twice.

evamill:
@hmcginnis My sense is that, while docs can go out any time, we especially want to line things up for the EOP releases. As that happened yesterday, you are now in the next cycle--it's one where we want to be sure certain docs go out on time (stuff for new features). Need to know what can possibly wait until after the Dec 23 merge, and it looks like this may be one of those.

terra4383:
Reviewed from a formatting/general copy edit standpoint. @evamill FYI it will be easy enough for me to merge this at the same time as the Java PR if it's all the same to you.

PR #822, opened 2019-12-06
Dotnet 980 net core installation reorder

m0nkey653:
I think the links at the top of the install page are auto-generated based on the docs there, so we didn't include them in the first section.

### DIFF
--- a/content/installation/netcore/install/NetCoreInstallation.md
+++ b/content/installation/netcore/install/NetCoreInstallation.md
@@ -10,7 +10,7 @@ tags: "installation agent .NET Core windows linux"
 To install the .NET agent, complete the following steps.
 > Note: The installation process for self-hosted applications also applies to IIS Express users.
 
-## Step 1 - Download Contrast Agent
+## Step One - Download the Contrast .NET Core Agent
 
 <details><summary><b>Download From Contrast UI</b></summary>
 
@@ -22,7 +22,7 @@ To install the .NET agent, complete the following steps.
 * On the web server, place the downloaded configuration file in a directory that your applications have sufficient permissions to access.
 </details>
 
-## Step 2 - Run the Application With Contrast
+## Step Two - Run the Application With Contrast .NET Core Agent
 
 To enable the .NET Core agent on your application, you must set the following environment variables on your application's process.
 Use the following table to find the correct CORECLR_PROFILER_PATH setting for the given architecture:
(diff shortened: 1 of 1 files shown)


m0nkey653:
Once @ldemidov reviews, I'll add UX to the PR.

content/installation/netcore/install/NetCoreInstallation.md
```diff
@@ -5,56 +5,38 @@ description: "Contrast .NET Core agent installation instructions"
 tags: "installation agent .NET Core windows linux"
 -->
 
-## The Basics
+## Install Overview
 
-To install the .NET agent, complete the following steps. The installation process for self-hosted applications also applies to IIS Express users.
+This document details the basic install of the .NET Core agent. The installation process for self-hosted applications also applies to IIS Express users.
```
evamill: Can you explain the change? The deleted text is actually better style than the change. Is it incorrect to say:

To install the .Net Core agent, complete the following steps. Note: the installation process for self-hosted applications also applies to IIS Express.

content/installation/netcore/install/NetCoreInstallation.md
```diff
@@ -5,56 +5,38 @@ description: "Contrast .NET Core agent installation instructions"
...
 
+## Step 1 - Download Contrast Agent
```
evamill: Style. More like:
Step One: Download the Contrast .Net Core agent

content/installation/netcore/install/NetCoreInstallation.md
```diff
@@ -5,56 +5,38 @@ description: "Contrast .NET Core agent installation instructions"
...
+
+<details><summary><b>Download the Contrast .NET Core Agent From Contrast UI</b></summary>
```
evamill: Redundant. This repeats the header. Should remove it.

content/installation/netcore/install/NetCoreInstallation.md
ldemidov: Yea, good point.  Fixed as you requested.

content/installation/netcore/install/NetCoreInstallation.md
ldemidov: updated

content/installation/netcore/install/NetCoreInstallation.md
ldemidov: I changed the text to say "Download From Contrast UI".  
We want to keep this text as hideable/accordion to keep the page smaller and more readable.  In the near future, we will add another option here for Nuget.org downloads.

PR #826, opened 2019-12-10
TS 1076 Update Monitor Attacks page

gdvdsn-contrast:
With the introduction of `Suspicious` as a new attack status, we need to update descriptions in the Attacks > Monitor Attacks page.


gdvdsn-contrast:
Updates to the docs have been published: https://docs.contrastsecurity.com/user-attacks.html#monitor.

PR #828, opened 2019-12-11
CONTRAST 39095 Change Jira password to token

gdvdsn-contrast:
This is a quick update to change "password" to "API token" in the Jira Integration.

Our UI refers to "key", so the screenshot doesn't quite match the language. But that will change eventually.


gdvdsn-contrast:
This content has been published and can be found here: https://docs.contrastsecurity.com/admin-orgintegrations.html#jira.

PR #841, opened 2020-01-06
UX-229 Vuln Mgmt - update 'auto-verified'

gdvdsn-contrast:
There are two instances of 'auto-verified' that need to be updated to 'auto-remediated'.


gdvdsn-contrast:
Updates made to: https://docs.contrastsecurity.com/admin-policymgmt.html#vuln-mgmt.

PR #842, opened 2020-01-09
Update Tomcat.md for potential Windows specific issues

pprofili-contrast:
The quotes in the environment variable SET command for Windows were misplaced - if they are placed as they were they become a literal part of the variable which leads to issues if CATALINA_OPTS is not already set to a valid value.  Using setenv.bat is by far the better option for Windows.


pprofili-contrast:
Anyone looking at these?

PR #855, opened 2020-02-06
Add SQLite to Java Supported Technologies

zirons1:


zirons1:
Kristen,

Each of those libraries, DB2, MySQL, and SQLite, all utilize a JDBC driver. So we're saying we support the JDBC drivers of these three data base libraries. Not that we support 4 drivers :)

kmckeedocs:
Thank you, this is how I learn! Consider it approved.

PR #859, opened 2020-02-11
Update python MySQL supported clients

danilito19:
See https://contrast.atlassian.net/browse/PYT-434 
We aren't supporting MySQLdb but do have two other clients.

See https://contrast.atlassian.net/browse/DOCS-31


lazorchakp:
I made a Contrast github account a while ago but I'm abandoning it in favor of my personal one. That's why there are two of me :)

PR #876, opened 2020-03-03
3.7.1 release notes

kmckeedocs:
Please add comments to this review by 3pm Pacific Time today. Comments can be left here or in [Google Docs](https://docs.google.com/document/d/1w4P-RGx8NdBfCnpH_zx9RVtgCSrS-9QaV6WuVNGIb20/edit?usp=sharing)

### DIFF
--- a/content/release/ReleaseNotes371.md
+++ b/content/release/ReleaseNotes371.md
@@ -1,0 +1,88 @@
+<!--
+title: "Contrast 3.7.1 - February 2020"
+description: "Contrast 3.7.1 February 2020"
+tags: "3.7.1 February Release Notes"
+-->
+
+## Release date:
+March 3, 2020
+
+### New features and improvements
+
+**LDAP-based Auto Group Provisioning**
+Customers can now leverage their internal groups to provision or deprovision users within Contrast groups.  When users authenticate to Contrast using LDAP, the user is automatically mapped to Contrast groups that share the same name as LDAP groups. Users that belong to a Contrast group, that are not in the LDAP group are dropped from the Contrast group. You can now use auto group mapping based on either LDAP or SAML (the SAML-based feature was released in 3.6.7).
+
+**Ghostcat CVE-2020-1938**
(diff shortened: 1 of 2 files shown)


content/release/ReleaseNotes371.md
```diff
@@ -0,0 +1,88 @@
...
+.NET Core: 2.1, 2.2, 3.0, 3.1
+
+#### Bug fixes:
+* Observed routes were not reported if the .NET agent was configured to use a proxy for communication with Contrast.
+* .NET agent could stop using the session ID specified via configuration after a period of time.
+* The `x-www-form-urlencoded` parameters could be reported as parameter keys.
+* .NET Core agent would not change Protect rule modes after initialization.
+
+#### New features and improvements:
+Added new gadgets for Protect Untrusted Deserialization and a session timeout rule for .NET Core. Most significantly, this release adds support for .NET Core 3.0 and 3.1.
```
DanFiedler: We should find a way to raise the prominence for adding support for .NET core 3.0 and 3.1. This is definitely the most important part of the .NET Core agent release.

content/release/ReleaseNotes371.md
```diff
@@ -0,0 +1,88 @@
...
+
+**Agent versions released this month:**
+NET Framework: 20.2.1, 20.2.2, 20.2.3 
+.NET Core: 1.2.0, 1.2.1, 1.2.2
+
+**Language versions supported:**
+.NET Framework: 3.5, 4.0, 4.5, 4.5.1, 4.5.2, 4.6, 4.6.1, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8
+.NET Core: 2.1, 2.2, 3.0, 3.1
+
+#### Bug fixes:
```
DanFiedler: We should find a way to rephrase the bug fixes to be positive. 

e.g. 
* Observed routes will now be reported correctly in cases where a non-system wide HTTP proxy was used.
* A session ID specified via configuration will now be used for the entire duration of that agent's deployment.
* Values for `x-www-form-urlencoded` parameters will no be correctly reported as parameter values rather than parameter keys. 
* The .NET Core agent now supports changing Protect rule modes at runtime.

kmckeedocs: Noted, however at this late hour, I'm going to follow the style guide and look into making a style update across the board if needed.

content/release/ReleaseNotes371.md
kmckeedocs: Resolved.

PR #881, opened 2020-03-09
Update ServiceInstall.md

tywkeene:
Centos 5 is end of life.


ao10:
I think this is fine, but @Avuxo would you mind confirming?

Avuxo:
LGTM. I don't see any reason why this should be a problem assuming product greenlit sunsetting REHL 5 & 6.

PR #884, opened 2020-03-10
JAVA-817 :pencil: usage of Java Agent with modules

argonotix:
Adds documentation on how to use the Java Agent with modules.


gdvdsn-contrast:
https://docs.contrastsecurity.com/installation-javaguides.html#modules

PR #903, opened 2020-04-03
Remove beta language about assess from Python docs

drdavella:
As of version 2.8.0 of the agent, Assess is no longer in Beta, so we need to remove this language from the docs.

### DIFF
--- a/content/installation/python/PythonOverview.md
+++ b/content/installation/python/PythonOverview.md
@@ -6,9 +6,8 @@ tags: "installation python agent overview"
 
 The Contrast Python agent provides runtime protection of Django, Flask and Pyramid web applications.
 
-Version 2.3.0 of the agent introduces beta Assess functionality. This provides the ability to find vulnerabilities in applications before they can be exploited.
-
->**Note:** The beta Assess feature can be licensed in **alpha** and **eval** instances of the Contrast UI.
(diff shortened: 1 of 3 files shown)


content/installation/python/PythonOverview.md
```diff
@@ -6,9 +6,7 @@ tags: "installation python agent overview"
 
 The Contrast Python agent provides runtime protection of Django, Flask and Pyramid web applications.
 
-Version 2.3.0 of the agent introduces beta Assess functionality. This provides the ability to find vulnerabilities in applications before they can be exploited.
-
->**Note:** The beta Assess feature can be licensed in **alpha** and **eval** instances of the Contrast UI.
+Version 2.8.0 of the agent introduces Assess functionality. This provides the ability to find vulnerabilities in applications before they can be exploited.
```
kmckeedocs, line 9: Address user directly, simplify. Does this wording work?

With Python agent version 2.8.0 you can find application vulnerabilities with Assess before they can be exploited.

drdavella: Yes that sounds great! Would it be okay to make say "Starting with...." just to be clear that it's a new feature?

kmckeedocs: Yes

